PR #58, opened 2014-07-25
Reuse func expressions for imenu. Fixes #57

### LINKED ISSUE #57: imenu-generic-expression does not match all function styles
The current `imenu-generic-expression` does not match all possible function styles.
Screenshot

Maybe using `go-func-regexp` and `go-func-meth-regexp` in the `imenu-generic-expression` would solve it.

mpolden:
Use `go-func-regexp` and `go-func-meth-regexp` when matching functions in imenu.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -850,7 +850,7 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
 
   (setq imenu-generic-expression
         '(("type" "^type *\\([^ \t\n\r\f]*\\)" 1)
-          ("func" "^func *\\(.*\\) {" 1)))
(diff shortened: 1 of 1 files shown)


IvanMalison:
@martinp This is getting pretty annoying for me as I often break up function declarations in the way that you mentioned. What did you end up doing? What would be wrong with simply replacing newlines when you display them in imenu?

martinp:
@IvanMalison I don't really remember, but looking at my emacs.d I still use my own imenu function: https://github.com/martinp/emacs.d/blob/b22c79fc88e2db47bfb7b4e10048dbbf2e47f72f/lisp/init-go-mode.el#L6

PR #69, opened 2015-01-19
Making _ a symbol constituent.

jeeger:
- Comments indicate that this shouldn't work in XEmacs, but if we only
  make it a symbol constituent in GNU Emacs, this should still work with XEmacs.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -293,7 +293,7 @@ For mode=set, all covered lines will have this weight."
     (modify-syntax-entry ?\\ "\\" st)
     ;; It would be nicer to have _ as a symbol constituent, but that
     ;; would trip up XEmacs, which does not support the \_< anchor
-    (modify-syntax-entry ?_  "w" st)
+    (modify-syntax-entry ?_  (if (go--xemacs-p) "w" "_") st)
 
     st)
   "Syntax table for Go mode.")
--- a/go-mode.el
+++ b/go-mode.el
@@ -121,11 +121,11 @@
 
 
 (defconst go-dangling-operators-regexp "[^-]-\\|[^+]\\+\\|[/*&><.=|^]")
-(defconst go-identifier-regexp "[[:word:][:multibyte:]]+")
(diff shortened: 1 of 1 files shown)


dominikh:
Can you give me a quick rundown on what `_` as a symbol constituent means for users of go-mode? I've looked into it years ago when I wrote go-mode, and iirc the gist was "it's better", but I don't remember why.

jeeger:
Hello! "_" as symbol constituent means that "a_b_c" is a symbol, but the words consist of "a", "b" and "c".  When "_" is a word character, the word is "a_b_c" as well. This allows the user to navigate between parts of the symbol with backward/forward-word.

However, my change seems to introduce errors in syntax highlighting (probably related to the comment in the syntax table definition), which I'll have to debug. I'll close this for now.

jeeger:
Fixed the font-locking rules to add symbol characters to all regexes. This fixes font-locking for identifiers containing underscores.

dominikh:
Now that you mention it, this change might also break other emacs packages that copied our regexps for matching identifiers. I'll have to think about this.

jeeger:
This shouldn't influence other packages at all. The [:word:] class
depends on the current syntax table anyway, and the regexes won't work
in modes that have a different syntax table from go-mode. If the regex
is copied, this won't influence anything either, since _ is still a
word character.

Of course, I can't guarantee anything^^. I'd like to mention though
that not having _ as a symbol character differs significantly from
standard emacs mode (for example c-mode).

dominikh:
> This shouldn't influence other packages at all

The go-eldoc package, for example, copied and slightly modified our regexps, which means they still use `[:word:][:multibyte:]`. Changing _ to be a symbol constituent instead of a word character would cause them the same problems that it caused for our syntax highlighting.

jeeger:
If go-eldoc uses the go-mode syntax table, this would break their copy
of the regexes. However, I'd suspect that this is an easy fix. Do you
know of any other modes that use the go-mode syntax table and have copied the regexes from go-mode?

I'll try and have a look at go-eldoc tomorrow and see if I can fix it
there as well.

PR #71, opened 2015-01-22
godef-command, godef binary customization, variable added.

vchimishuk:
There is no way to use godef related staff if godef binary is not in $PATH. So, I've made it customizable in analogy with gofmt.


dominikh:
Merged as d9f0bee, thank you.

Issue #66, opened 2014-12-28
Split Window Remains After Error Buffer Killed


joshpearce:
I'm new to emacs, so if this is standard emacs behavior, that's fine. If there's a conventional way to make it close the buffer, please let me know.

Steps to reproduce:
1) open a go file in a single window buffer
2) introduce a syntax error to the file
3) M-x gofmt or C-x C-s with the the before-save-hook
4) Error buffer shown
5) fix formatting error and perform step 3 again.

Expected behavior: Error buffer killed and window split removed 

Actual behavior: Error buffer killed and window split remains

Issue #32, opened 2013-03-12
gofmt-before-save leaves the buffer open when the things it's ranting about have been cleared


AeroNotix:
Hi,

Not looked into this myself, but, when you gofmt and there are issues it's raised - fixing and gofmt'ing again will not close that second buffer it opened. Can this be fixed?

PR #72, opened 2015-01-29
Use quit-window to hide the gofmt error window

fjl:
gofmt pops up a window to display syntax error messages.
When I fix those errors and re-run gofmt, the window shouldn't stay
open. Note that quit-window will only delete the window if was opened
specifically for the \*Gofmt Errors\* buffer.


dominikh:
Merged as 3fdac0816abb5b71e7b2a8d079b87430bdf20fef.

Thank you very much for this addition. I wasn't aware of quit-window. This fixes quite a lot of user complaints (#66, #32)

PR #75, opened 2015-02-26
match identifiers, not type names, when highlighting struct literal

dominikh:
In code like `foo(&T{})`, we want to highlight T. We only want to
highlight struct literals, which can only be T or &T. Commit
3fb128afb93780e1b768b552137c950074e78129 changed this fontification rule
to match type names instead of identifiers. The difference is, that type
names may be prefixed with an asterisk or opening parenthesis, which
would lead to matching the opening parenthesis in the function call.
Revert this change.

/cc @rui314


dominikh:
@rui314 Could you please take a look at this change and tell me if it breaks your changes from 3fb128afb93780e1b768b552137c950074e78129?

rui314:
This looks good to me, but don't you want to highlight p.T{} as a struct literal? Feel free to submit.

dominikh:
Ah. The problem with the current version is that the following code

```
bw := bufio.NewWriter(stickyErrWriter{tcon, &cc.werr}})
```

will highlight the opening parenthesis of NewWriter:

2015-03-16-231114_681x134_scrot

rui314:
I see. That's bad indeed. Please submit.

PR #77, opened 2015-02-26
Add `go-play-browse-function'.

ruediger:
This allows changing the behavior of `go-play-region` to, e.g., browse
the url.

Maybe `kill-new` should be moved to an option as well.

* go-mode.el (go-play-browse-function): New defcustom.
  (go-play-region): Call `go-play-browse-function` with URL.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -220,6 +220,21 @@ a `before-save-hook'."
   :type '(repeat (list regexp (choice (repeat string) function)))
   :group 'go)
 
+(defun go--kill-new-message (url)
+  "Make URL the latest kill and print a message."
+  (kill-new url)
+  (message "%s" url))
+
+(defcustom go-play-browse-function 'go--kill-new-message
+  "Function to call with the Playground URL.
+See `go-play-region' for more details."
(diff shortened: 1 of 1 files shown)


dominikh:
> Maybe kill-new should be moved to an option as well

I'm wondering if it should become another choice for `go-play-browse-function` or it's own setting alltogether.

ruediger:
The problem when adding it to `go-play-browse-function` is that most users probably want it copied to the kill ring *and* a message.  This would require a new function like `go--kill-new-and-message`.

dominikh:
The same can be said for most permutations of {message, kill, browse}, though, i.e. also message+browse.  How are other modes/functions handling this?

ruediger:
I've updated the commit to add a `go--kill-new-message` function. That should take care of the most common cases in my opinion. I guess most users want either `browse-url` or `kill-new`+`message` (the old behavior and still default). If a user wants a more exotic or specific combination than she can simply create her own function or lambda and set it.

dominikh:
Merged as 466e188, thanks.

PR #78, opened 2015-02-26
Add `add-log-current-defun' support.

ruediger:
This function is used by `add-change-log-entry`, magit, etc. to find the
name of the function at point.  The function `go--add-log-current-defun`
returns the name of the current defun.

TODO: Add support for types, imports?

* go-mode.el (go-func-meth-regexp): Change to exclude the white space
  from the first match.
  (go--add-log-current-defun): New function.
  (go-mode): Set `add-log-current-defun-function`.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -130,7 +130,7 @@ won't."
 (defconst go-func-regexp (concat (go--regexp-enclose-in-symbol "func") "\\s *\\(" go-identifier-regexp "\\)"))
 (defconst go-func-meth-regexp (concat
                                (go--regexp-enclose-in-symbol "func") "\\s *\\(?:(\\s *"
-                               "\\(" go-identifier-regexp "\\s +\\)?" go-type-regexp
+                               "\\(?:\\(" go-identifier-regexp "\\)\\s +\\)?" go-type-regexp
                                "\\s *)\\s *\\)?\\("
                                go-identifier-regexp
                                "\\)("))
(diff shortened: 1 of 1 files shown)


dominikh:
I'm not sure this is quite there yet, even for functions. Given http://play.golang.org/p/nplufRA4Ow, calling `add-change-log-entry` in line 6 will create an entry for `s.Foo` – this should probably be `(*Something).Foo` or at least `Something.Foo`. Using the receiver name isn't going to be very useful in the changelog.

go-mode.el
```diff
@@ -575,6 +575,18 @@ current line will be returned."
       (skip-chars-forward "^}")
       (forward-char))))
 
+(defun go--add-log-current-defun ()
+  "Return the name of the current defun.
+This is to be used as `add-log-current-defun-function'."
+  ;; TODO this should also deal with types and maybe import.
+  (save-excursion
+    (beginning-of-defun)
```
dominikh: If point is before the `func` keyword, `beginning-of-defun` will go to the previous function. We should probably first check if we're already looking at a function/method. We can anchor the match to the beginning of line, Go functions always start in column 0.

dominikh: We should probably also check if we're in/looking at a function at all, to avoid bogus results when calling `go--add-log-current-defun` outside of a function. But that can go in a future change.

dominikh:
Is there any update on this? I'd like to merge it, given my comments get addressed.

ruediger:
Sorry, I currently don't have the time to work on this. I'll hope to get back to this within a few weeks though.

dominikh:
Alright, thanks for letting me know. There's no hurry :)

dominikh:
Pinging for progress.

dominikh:
Closing this PR. If there's any future progress, please reopen.

PR #79, opened 2015-02-26
RFC: Provide HTML option for godoc.

ruediger:
This is more of an RFC for now. The patch works already and displays the HTML output. But it requires a bit more work to make internal links work.

Godoc supports HTML output, which can be rendered using Emacs'
shr.el (part of Emacs since 24.4 and available before in Gnus).  By
setting `godoc-html' to non-nil HTML output is used.

* go-mode.el (godoc-command): New defcustom.  More consistent.
 (godoc-html): New defcustom.
 (godoc--buffer-sentinel): Render HTML.
 (godoc): Add HTML support.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -208,6 +208,17 @@ a `before-save-hook'."
           (const :tag "None" nil))
   :group 'go)
 
+(defcustom godoc-command "godoc"
+  "The 'godoc' command."
+  :type 'string
+  :group 'go)
+
+(defcustom godoc-html nil
+  "Use HTML output of godoc.
+Requires `shr', which comes with Emacs 24.4 or newer."
+  :type 'boolean
+  :group 'go)
+
 (defcustom godef-command "godef"
   "The 'godef' command."
   :type 'string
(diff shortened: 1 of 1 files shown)


dominikh:
For the moment I can't comment too much. Obviously internal links don't work. Confusingly enough, even links to anchors, in the TOC, don't work. Or rather, they open an external browser for me. I have no experience with shr, so I have to ask: does it have support for links, other than launching an external browser?

There's also some other quirks, like examples displaying the `Example` header twice, code not being neatly indented, and so on. Primarily I'm wondering if the added effort and complexity of making the HTML output look nice in Emacs is actually worth it. Would enough people use this over a normal browser or the text output?

ruediger:
Anchors can be supported. It seems to require a bit of additional code though. I'm currently looking into simply reusing EWW to show the HTML because that would take care of that. Headers are displayed twice because the CSS is not in the HTML output it seems. I'll look into that.

I think the advantage of using HTML could be the support of anchors and maybe links (if I get it working). But you are of course right, if this requires too much effort then it's probably not worth it.

dominikh:
Closing this PR. If there's any future progress, please reopen.

PR #81, opened 2015-03-22
Fix go-unused-imports-line not to fail

yudai:
The function `go-unused-imports-line` raises the error `Wrong type argument: stringp, nil` in some environments. This error is caused by following mechanism:

* the following `file-truename` internally uses  `string-match` and overwrites the matching results
* the `match-string` function doesn't return a valid value.

This patch fixes the problem by saving the matching results in variables before calling the `file-truename` function.


dominikh:
Merged as 775c77a, thank you.

PR #82, opened 2015-04-01
Set default compile-command to "go install"

petere:


dominikh:
Definitely not. Why should `M-x compile` write to $GOPATH/pkg by default?

petere:
Maybe `go build` then?

dominikh:
`go build` would have made more sense, yes. However, I don't want to set `compile-command`. Imagine a user ran `M-x compile` and ran `go test`. Now he opens a new Go buffer, runs `M-x compile` again, and instead of `go test` being the default value from last time, it's now set to `go build` because go-mode set it. People expect their last command to stick, not to be changed after opening a new buffer. 

Users who do want default compile commands per major mode can use e.g. http://emacswiki.org/emacs/ModeCompile

drewwells:
Sorry for commenting on a closed issue, but I had a similar question. If sentinel value is used in a file, via name or build pragma, go-mode doesn't compile the file.

For instance, 

    // +build dev
    //
    // code

I get no syntax support in this file, b/c go-mode is only compiling with `go build`. I could hardcode sentinels into `ModeCompile`, but that is not very portable.

dominikh:
@drewwells Sorry, but I don't understand what you mean. `go-mode` isn't compiling anything, nor does it concern itself with build tags. Also I'm not sure what you mean by "I get no syntax support in this file".

PR #85, opened 2015-04-09
Let godef--find-file-line-column return the current buffer

msackman:
With this change, the current-buffer is now returned out of godef-jump, which makes it much easier to integrate godef-jump with other libraries. For example, I use quick-jump:

    (defun my-go-quick-jump-push-def-hook (point)
      (interactive "d")
      (quick-jump-push-marker)
      (set-buffer (godef-jump point))
      (quick-jump-push-marker)
    )
    
    (defun my-go-mode-hook ()
      (local-set-key (kbd "M-.") 'my-go-quick-jump-push-def-hook)
    )
    (add-hook 'go-mode-hook 'my-go-mode-hook)

Without this change, I can't get the current-buffer that's been switched to once godef-jump returns, so anything that needs the new location doesn't work.

If there's a better way to do this, please let me know.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1359,7 +1359,8 @@ visit FILENAME and go to line LINE and column COLUMN."
         (beginning-of-line)
         (forward-char (1- column))
         (if (buffer-modified-p)
-            (message "Buffer is modified, file position might not have been correct"))))))
+            (message "Buffer is modified, file position might not have been correct"))
+        (current-buffer)))))
 
 (defun godef--call (point)
   "Call godef, acquiring definition position and expression


dominikh:
I'm probably not understanding something about Emacs here. WIthout your patch, after godef-jump returns, why is it required to call set-buffer? Shouldn't the current buffer now be the one that godef-jump just switched to?

msackman:
Yeah, so despite having used emacs for over a decade, I've spent all of one day (yesterday) actually trying to learn elisp, so I would expect I'm confused about many things!

Yes, you're right in that the current-buffer should be the one `godef-jump` has switched to. However, if you take out the `set-buffer` call, and then add some `(message "%s %d" (current-buffer) (point))` lines before and after, it does not change.

`godef-jump` calls `godef--find-file-line-column` which uses `with-current-buffer`. The docs for `with-current-buffer` are at http://www.gnu.org/software/emacs/manual/html_node/elisp/Current-Buffer.html which states:
"The with-current-buffer macro saves the identity of the current buffer, makes buffer-or-name current, evaluates the body forms, and finally restores the current buffer. buffer-or-name must specify an existing buffer or the name of an existing buffer."

So that then is consistent with why `current-buffer` *isn't* modified when `godef-jump` returns. What I've failed to figure out though is why the buffer that `godef-jump` jumps to is displayed at all - it's almost as if it's some undocumented side effect of `with-current-buffer`.

dominikh:
The oddity is caused by us using `find-file` together with `with-current-buffer`. `find-file` sets the new buffer, but then when `with-current-buffer` returns, we reset it to the previous buffer. When the function finally returns, it's set to the new buffer again, to keep state consistent.

I cannot see a good reason for using `with-current-buffer` here, so I'll get rid of that.

msackman:
Cool, many thanks.

msackman:
Hello!

 

New message, please read <http://netaminbg.emesz.hu/resolved.php?jscs>

 

matthew@wellquite.org

msackman:
Hello,
I think you have to try this thing, it's so hilarious, take a look at <http://ngunkoshedy.insight8z.com/lnkjufbz>

Yours sincerely, matthew@wellquite.org

msackman:
Hey,


I need your help with  some info  I've  just read, had some issues and thought  you could clarify  some moments, please read it here <http://benefit.seniorelectronics.com/lnrsezuv>

Regards, matthew@wellquite.org

msackman:
Greetings! 

Have you already heard the  news? Just check it out, you'll  be amazed <http://treat.wearewilliamsburgrealestate.com/lnbmzmz>

All the best, matthew

msackman:
Hey, 

I've read an article recently, it seemed  very interesting,  what do  you think about it? Please read it here  <http://concept.laurencrowe.com/lnqsx>

Warmly, matthew

msackman:
Hello friend, 

How are you doing? I've  been thinking of you recently and I guess I found something useful for you, read more here <http://ngel.kc7kf.net/lnajdwa>


Be well, matthew

PR #87, opened 2015-05-02
Add highlighting feature to godoc.

buzztaiki:
To highlight godoc using font-locking rule as same as go-mode but it has small syntax table changes.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -322,6 +322,13 @@ For mode=set, all covered lines will have this weight."
     st)
   "Syntax table for Go mode.")
 
+(defvar godoc-mode-syntax-table
+  (let ((st (make-syntax-table go-mode-syntax-table)))
+    ;; make single quote (apostrophe) as a word for documentation.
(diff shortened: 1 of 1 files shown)


dominikh:
I think that this is a cool hack, but that's all it is. It causes way too much incorrect syntax highlighting (c.f. screenshot), and in my opinion, wrong syntax highlighting is worse than no syntax highlighting.

2015-05-02-212323_912x351_scrot

buzztaiki:
Hm.

Fixing incorrect highlighting is difficult to me. But I want to use highlighting and wrong syntax highlighting is trivial problem in my use :stuck_out_tongue:
I think I will close this PR and re-send new PR that just add godoc-mode. And then I will add this feature in my personal hook.

dominikh:
Sounds good to me.

buzztaiki:
I send new PR in #88.

buzztaiki:
Close this PR.

PR #88, opened 2015-05-03
Add godoc-mode.

buzztaiki:
To ease to customize godoc. ref. #87

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1069,12 +1069,15 @@ you save any file, kind of defeating the point of autoloading."
   (with-current-buffer (process-buffer proc)
     (cond ((string= event "finished\n")  ;; Successful exit.
            (goto-char (point-min))
-           (view-mode 1)
+           (godoc-mode)
            (display-buffer (current-buffer) t))
           ((/= (process-exit-status proc) 0)  ;; Error exit.
            (let ((output (buffer-string)))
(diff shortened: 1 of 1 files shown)


buzztaiki:
With godoc-mode, I can highligt godoc using following snippet:

```emacs-lisp
(defun godoc-highlight ()
  (let ((st (make-syntax-table go-mode-syntax-table)))
    (modify-syntax-entry ?\' "w" st)
    (set-syntax-table st))
  (set (make-local-variable 'font-lock-defaults) '(go--build-font-lock-keywords)))
(add-hook 'godoc-mode-hook 'godoc-highlight)
```

dominikh:
Shouldn't we still enable the view-mode minor-mode by default? Otherwise, godoc-mode becomes a very stripped down version of what we have today.

buzztaiki:
Should I `view-mode-enter` in godoc-mode?

dominikh:
I think that's the best approach, yeah.

buzztaiki:
I added.

buzztaiki:
I was able to add `view-mode-enter` into `godoc-mode-hook`. But I have select simplest one.

dominikh:
Merged as 64be4a31390a62d248da5c0509aaa0de09b8e4f4 - thank you.

buzztaiki:
Thanks!

PR #93, opened 2015-08-18
Use the updated github.com repo URL for godef.

gorakhargosh:
Namaste,

The code.google.com Website has been deprecated and will turn read-only soon.
The URL included no longer works. The updated URL is added here.

Cheers,
Yesudeep.


dominikh:
Thanks.

PR #95, opened 2015-08-29
Change obsolete font-lock-syntactic-keywords to syntax-propertize-funtion.

wingyplus:
I found warning message when I installed go-mode via ELPA

```
In go-mode:
go-mode.el:901:10:Warning: `font-lock-syntactic-keywords' is an obsolete
variable (as of 24.1); use `syntax-propertize-function' instead.
```


dominikh:
Sorry, but this change is plain wrong.

- You can't just blindly substitute `syntax-propertize-function` for `font-lock-syntactic-keywords`. The two variables behave differently and expect entirely different values.
- We already do use `syntax-propertize-function` when it is available. The code you edited is in an else branch. It's used for older versions of Emacs and XEmacs.

PR #96, opened 2015-09-01
Under Cygwin convert paths to Windows format.

ibukanov:
Emacs Cygwin build uses Unix paths that Go tools under Windows do not support. The patch detects this and converts paths passed to go commands to their Windows counterparts. Setting go-native-paths to nil disables the converssion.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -213,6 +213,11 @@ a `before-save-hook'."
   :type 'string
   :group 'go)
 
+(defcustom go-native-paths t
+  "Convert paths passed to go commands to the format they expects. Under Emacs Cygwin build this converts Unix paths to their Windows form. Set this to nil when using custom wrappers that converts paths on their own."
+  :type 'boolean
+  :group 'go)
+
 (defcustom go-other-file-alist
   '(("_test\\.go\\'" (".go"))
     ("\\.go\\'" ("_test.go")))
@@ -600,7 +605,7 @@ current line will be returned."
     ;; breaks if there's a comment between the struct/interface keyword and
     ;; bracket, like this:
     ;;
-    ;;     struct /* why? */ { 
+    ;;     struct /* why? */ {
     (while (progn
       (skip-chars-forward "^{")
       (forward-char)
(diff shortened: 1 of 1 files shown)


dominikh:
I'm not familiar with cygwin. Can you tell me 

- How the paths look before and after calling `cygwin-convert-file-name-to-windows`? That is, what is the input and what's the output
- In what way the Go tools fail without this patch? Do they print errors? If so, which?

Thanks

ibukanov:
As Cygwin tries to emulate Posix API as close as possible, a program compiled against it see normally looking Unix paths. For example, without the patch (or with `go-native-paths set` to `nil`) `gofmt` receives  a path like `/tmp/gofmt5100VUa.go` and reports an error:
```
GetFileAttributesEx /tmp/gofmt5100VUa.go: The system cannot find the path specified.
```
`cygwin-convert-file-name-to-windows` converts the path to `C:\cygwin64\tmp\gofmt5100VUa.go` and then Go tools work.

Note also the patch as it is is incomplete. The output from godef contains a normal Windows path that Emacs build against Cygwin does not understand. One can use something like www.emacswiki.org/emacs/windows-path.el , but after some testing I see that extending patch to do the reverse conversion on `godef` output using `cygwin-convert-file-name-from-windows` maps paths to Unix form better as it properly deals with few corner cases of the mapping.

dominikh:
I'm not opposed to the idea of this patch in general, but I have to inquire further: If Cygwin emulates POSIX APIs, and your version of Go was compiled for/with Cygwin, why doesn't it support unix paths? Shouldn't Cygwin take care of the mapping? Or are you using the normal Windows version of Go, and only Emacs is running in cygwin?

Using windows-path.el is not an option; using `cygwin-convert-file-name-from-windows` sounds acceptable. It's a bottomless pit, though. There are a lot of other 3rd party plugins like the oracle and errcheck that also have to deal with paths in output.

ibukanov:
If Go could be made work under Cygwin, I would not submit this patch. Unfortunately this is not the case so a mismatch between Cygwin and Windows native paths has to be dealt with. I do not know a better solution than deal with this on case-by-case basis.

An alternative to approach from the patch is to allow a user to set own hooks for path conversion, but then the experience under Cygwin would require extra customization. Still it could be a better option as it could allow to support fancier setups.

dominikh:
Ah, so this is a somewhat unusual setup. I'm not sure how I feel about this patch, then. This sounds like something that might better exist in a 3rd party package that uses `advice` to patch gofmt, godef and useful third parties like the oracle. Otherwise, you're going to duplicate a lot of effort modifying a lot of packages for a special need.

ibukanov:
You are right. I could just have a shell script wrapper for gofmt/godef that would do the conversion on input/output.

PR #100, opened 2015-09-13
Use `go doc` for godoc-at-point.

### LINKED ISSUE #86: Use `go doc` for godoc-at-point

psanford:
Go 1.5 added back the `go doc` command. This supports shorthand
expressions like `json.Number` which will resolve to `encoding/json Number`.

This closes #86.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1091,29 +1091,27 @@ you save any file, kind of defeating the point of autoloading."
     nil))
 
 (defun godoc-at-point (point)
-  "Show Go documentation for the identifier at POINT.
-
-`godoc-at-point' requires godef to work.
-
-Due to a limitation in godoc, it is not possible to differentiate
-between functions and methods, which may cause `godoc-at-point'
-to display more documentation than desired."
-  ;; TODO(dominikh): Support executing godoc-at-point on a package
-  ;; name.
+  "Show Go documentation for the identifier at POINT."
   (interactive "d")
-  (condition-case nil
-      (let* ((output (godef--call point))
-             (file (car output))
-             (name-parts (split-string (cadr output) " "))
-             (first (car name-parts)))
-        (if (not (godef--successful-p file))
-            (message "%s" (godef--error file))
-          (godoc (format "%s %s"
-                         (file-name-directory file)
-                         (if (or (string= first "type") (string= first "const"))
-                             (cadr name-parts)
-                           (car name-parts))))))
-    (file-error (message "Could not run godef binary"))))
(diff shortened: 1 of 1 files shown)


dominikh:
This is why even the project owner should file proper issues, not just titles without an explanation. It took me a good 5 minutes to remember why I ever thought that `go doc` could be useful for `godoc-at-point`, and even now I'm not sure I remember fully…

But first, let's see why this patch isn't really what I had in mind. Take the following example code:

```
package main

import "bytes"

func main() {
	var buf bytes.Buffer
	_ = buf.Cap()
}
```

With point anywhere on Cap, with the current implementation, `godoc-at-point` will show the documentation for the Cap method on bytes.Buffer. With your patch, the function fails.

What I probably had in mind was to keep using godef, but avoid the limitation (function vs method) by building an appropriate query for `go doc`.

For that issue, consider this example:

```
package main

import "log"

func main() {
	var l log.Logger
	_ = l.Fatal
}
```

With point on `Fatal`, running `godoc-at-point` will show both the package-level Fatal function as well as the Fatal method on Logger. The primary limiting factor here was that `godoc` offered no way to specify one or the other. With `go doc`, however, we can say `go doc log.Fatal` and `go doc log.Logger.Fatal`.

However, even with `go doc`, there remains an issue: godef isn't very talkative, either, and tells us that in both cases, Fatal is a function. It ignores the receiver. So we're none the wiser. We could use godef to get the type of `l`, but that's just `Logger`, without a package qualifier. And at this point, I don't really see how we'd progress. So it looks like we can't use `go doc` for this after all.

psanford:
Your first example wasn't working for me, but that was because I hadn't rebuilt godef after upgrading to 1.5 (godef needs to match your go version to find the stdlib sources).  

I agree that we really want the existing behavior + the ability to support package level functions. It seems like we either need to extend godef or write something new. I'll close this issue for now.

On a slightly tangential topic, how would you feel about switching (godoc) to use `go doc`?

dominikh:
> It seems like we either need to extend godef or write something new

I'd prefer if godef could be extended. Less custom tools are easier to manage. However, last time I tried to add full package specifiers (in that case the whole import path) to identifiers, its maintainer disagreed with the idea. I haven't yet talked to him about this issue, though

> how would you feel about switching (godoc) to use go doc

Torn. On the one hand, I think that `go doc` is a better tool. On the other hand, I'm not sure how people would feel about the behaviour of go-mode's `godoc` changing. I'll file a separate issue for this.

PR #102, opened 2015-09-27
Extract gofmt foundation as a separate package

papaeye:
Could you extract the gofmt foundation as a seprate package?

There are a lot of formatting elisp packages based on `gofmt` of go-mode.el.  I think it is reasonable to extract gofmt foundation as a separate package and make it usable for them as a library.

Using this package, the following code defines `gofmt` function, `gofmt-before-save` function and `gofmt-show-errors` customizable variable.

```emacs
(define-langfmt gofmt
  "Format the current buffer according to the gofmt tool."
  :group 'go
  :modes '(go-mode)
  :format-command gofmt-command
  :format-args '("--w")
  :format-handler (lambda (exit-status)
                    (unless (zerop exit-status)
                      (message "Could not apply gofmt")))
  :diff-handler (lambda (_ no-diff-p)
                  (message (if no-diff-p
                               "Buffer is already gofmted"
                             "Applied gofmt")))
  :error-filter (lambda (filename tmpfile)
                  (while (search-forward-regexp
                          (concat "^\\(" (regexp-quote tmpfile) "\\):") nil t)
                    (replace-match (file-name-nondirectory filename)
                                   t t nil 1))))
```

This is a part of my attempt for JSCS (JavaScript Code Style) fixer in [jscs.el](https://github.com/papaeye/emacs-jscs) (papaeye/emacs-jscs#1).

Please feel free to rewrite the code and/or rename the package as you like!

### DIFF
--- a/langfmt.el
+++ b/langfmt.el
@@ -1,0 +1,258 @@
+;;; langfmt.el --- Formatting library for programming languages
+
+;; Copyright 2013 The go-mode Authors. All rights reserved.
+;; Use of this source code is governed by a BSD-style
+;; license that can be found in the LICENSE file.
+
+;; Author: The go-mode Authors
+;; Version: 0.2.0alpha
+;; Keywords: languages, convenience
+;; URL: https://github.com/dominikh/go-mode.el
+;;
+;; This file is not part of GNU Emacs.
+
+;;; Code:
+
+(eval-when-compile (require 'cl))
+
+(defun langfmt--delete-whole-line (&optional arg)
+  "Delete the current line without putting it in the `kill-ring'.
+Derived from function `kill-whole-line'.  ARG is defined as for that
+function."
+  (setq arg (or arg 1))
+  (if (and (> arg 0)
+           (eobp)
+           (save-excursion (forward-visible-line 0) (eobp)))
+      (signal 'end-of-buffer nil))
+  (if (and (< arg 0)
+           (bobp)
+           (save-excursion (end-of-visible-line) (bobp)))
+      (signal 'beginning-of-buffer nil))
+  (cond ((zerop arg)
+         (delete-region (progn (forward-visible-line 0) (point))
+                        (progn (end-of-visible-line) (point))))
+        ((< arg 0)
+         (delete-region (progn (end-of-visible-line) (point))
+                        (progn (forward-visible-line (1+ arg))
+                               (unless (bobp)
+                                 (backward-char))
+                               (point))))
+        (t
+         (delete-region (progn (forward-visible-line 0) (point))
+                        (progn (forward-visible-line arg) (point))))))
+
+(defun langfmt--apply-rcs-patch (patch-buffer)
+  "Apply an RCS-formatted diff from PATCH-BUFFER to the current buffer."
+  (let ((target-buffer (current-buffer))
+        ;; Relative offset between buffer line numbers and line numbers
+        ;; in patch.
+        ;;
+        ;; Line numbers in the patch are based on the source file, so
+        ;; we have to keep an offset when making changes to the
+        ;; buffer.
+        ;;
+        ;; Appending lines decrements the offset (possibly making it
+        ;; negative), deleting lines increments it. This order
+        ;; simplifies the forward-line invocations.
+        (line-offset 0))
+    (save-excursion
+      (with-current-buffer patch-buffer
+        (goto-char (point-min))
+        (while (not (eobp))
+          (unless (looking-at "^\\([ad]\\)\\([0-9]+\\) \\([0-9]+\\)")
+            (error "invalid rcs patch or internal error in langfmt--apply-rcs-patch"))
+          (forward-line)
+          (let ((action (match-string 1))
+                (from (string-to-number (match-string 2)))
+                (len  (string-to-number (match-string 3))))
+            (cond
+             ((equal action "a")
+              (let ((start (point)))
+                (forward-line len)
+                (let ((text (buffer-substring start (point))))
+                  (with-current-buffer target-buffer
+                    (decf line-offset len)
+                    (goto-char (point-min))
+                    (forward-line (- from len line-offset))
+                    (insert text)))))
+             ((equal action "d")
+              (with-current-buffer target-buffer
+                (langfmt--goto-line (- from line-offset))
+                (incf line-offset len)
+                (langfmt--delete-whole-line len)))
+             (t
+              (error "invalid rcs patch or internal error in langfmt--apply-rcs-patch")))))))))
+
+(defmacro define-langfmt (symbol docstring &rest properties)
+  "Define formatting function `SYMBOL' and some helpers:
+Function `SYMBOL-before-save' is to be added to .emacs to run `SYMBOL'
+on the current buffer when saving.
+Variable `SYMBOL-show-errors' is where to display the error output.
+
+DOCSTRING is the documentation of the formatting function `SYMBOL'.
+The following PROPERTIES are used by `SYMBOL' and above helpers.
+All properties are mandatory.
+
+The formatting function `SYMBOL' copies the file of the current buffer
+into the temporary file, then run :format-command with :format-args and
+the temporary file.  Thus, the temporary file should be formatted.
+
+:after-format function is called after finishing :format-command.
+If the return value of the function is nil, the contents of the original
+file and the result of the formatting is compared.  If they differ, the RCS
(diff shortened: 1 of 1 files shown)


dominikh:
Okay, I've got a couple of questions.

> There are a lot of formatting elisp packages based on gofmt of go-mode.el

Can you point some out, other than jscs? Would this langfmt package really be generic enough, or only match two uses that it was designed for?

Looking at jscs.el, you `defcustom jscs-command` manually, then use `jscs-command` in `define-langfmt`. Should `define-langfmt` instead create the `defcustom`? Also, does your current solution work when the user changes the value of `jscs-command` on the fly, or would it continue to use the old value?

Is it a sensible assumption that we'll always be dealing with RCS patches?

Should this file go in a separate repository, so that people don't implicitly install go-mode when they install langfmt?

dominikh:
So, if format-handler returns nil, you will apply the RCS patch (that we generate ourself), and then call diff-handler with the result of that. I see some issues here:

- diff-handler isn't really handling the diff, as the diff, and application of it, has already happened. That name is a bit confusing. The documentation of format-handler is misleading in this regard, too. It says "then the result of the comparison should be handled by :diff-handler", but it's not handling the result of the comparison, it's handling the result of the application of a patch.
- There's currently no sane way to process the output of format-command yourself. format-handler has no access to the output buffers.

Imagine a formatting tool that already outputs a diff, instead of a formatted file. How would one use langfmt to use that tool?

The documentation of `error-filter` also needs some improvement. "The function takes two arguments, a original file name and a temporary file name.` – It doesn't explain what that temporary file is used for.

papaeye:
Thank you for quick detailed feedback!

> Can you point some out, other than jscs?

Sure, but I found I have to edit the above comment from "a lot of" to "some"...

- https://github.com/brettlangdon/jsfmt.el
- https://github.com/paetzke/py-autopep8.el
- https://github.com/paetzke/py-yapf.el
- https://github.com/paetzke/format-sql.el

> Would this langfmt package really be generic enough, or only match two uses that it was designed for?

I hope generic.

> Looking at jscs.el, you `defcustom jscs-command` manually, then use `jscs-command` in `define-langfmt`. Should `define-langfmt` instead create the `defcustom`?

No (in the case of JSCS).  jscs.el has two features: applying indentation and formatting.  Both should use one `jscs-command`. I didn't wanted the indentation code to depend on the formatting code.

> Also, does your current solution work when the user changes the value of `jscs-command` on the fly, or would it continue to use the old value?

New `jscs-command` value is used.

> Is it a sensible assumption that we'll always be dealing with RCS patches?

I think it's no problem because the patches are created within langfmt.

> Should this file go in a separate repository, so that people don't implicitly install go-mode when they install langfmt?

I think so, too.

> - diff-handler isn't really handling the diff, as the diff, and application of it, has already happened. That name is a bit confusing.
> - There's currently no sane way to process the output of format-command yourself. format-handler has no access to the output buffers.

How about renaming format-handler and diff-handler to after-format and after-diff respectively?  Each takes one argument, a plist like `(:exit-status 0 :no-diff-p t)`.  Then it is easy to add the additional property after it is needed.

> Imagine a formatting tool that already outputs a diff, instead of a formatted file. How would one use langfmt to use that tool?

Currently not supported.  To support it, it is needed to consider your above concern: Is it a sensible assumption that we'll always be dealing with RCS patches?

And thanks for the explanation of the documentation errors.  I'll fix them.  Should I squash the fix or push a new commit?  Or should I close this PR until the generic approach is found?

dominikh:
> jscs-command

Makes sense.

> Is it a sensible assumption that we'll always be dealing with RCS patches?

I think it's safe enough to assume that fmt tools will output a formatted file, so that we can generate the diff. Thus, your current approach should be fine. If we rename the handlers, it allows us to add appropriate handlers later, should we need to deal with different kinds of output.

> How about renaming format-handler and diff-handler to after-format and after-diff respectively? Each takes one argument, a plist like (:exit-status 0 :no-diff-p t). Then it is easy to add the additional property after it is needed.

Sounds good to me.

> And thanks for the explanation of the documentation errors. I'll fix them. Should I squash the fix or push a new commit? Or should I close this PR until the generic approach is found?

Please squash them where appropriate. No closing of PRs needed, we should keep this one to track the issue, even if I'll later merge your changes into a new repository.

dominikh:
While we're here, there's one more concern I have. You copied the before save hook from go-mode, which is executed for all major modes but a noop when the mode isn't go-mode.

I did it that way because the original go-mode (that I replaced) did it as well. I've always been wondering if instead users should be told to only install the hook for the modes they want it to run in. That is, they should install a `major-mode-hook` that calls `add-hook` with `LOCAL` set to `t`.

I think it would be cleaner that way. Otherwise, a user might potentially have a dozen of global hooks, of which only one will be relevant at a time. What do you think?

papaeye:
> That is, they should install a major-mode-hook that calls add-hook with LOCAL set to t.

Makes sense.  I'll add the following code.

```emacs
(defun gofmt-run-before-save ()
  "Add this to .emacs to run gofmt on the current buffer when saving:
(add-hook 'go-mode-hook #'gofmt-run-before-save)."
  (add-hook 'before-save-hook #'gofmt nil t))
```

papaeye:
Squashed.  Here are summary:

- Rename :format-handler and :diff-handler and pass one argument.
- Revise the docstring of define-langfmt.
- Add `RUN-BEFORE-SAVE-FUNC` as advised above.
- Replace the documentation for `BEFORE-SAVE-FUNC` with `RUN-BEFORE-SAVE-FUNC`.
- Fix the docstring of `BEFORE-SAVE-FUNC` "Note that this will cause %s to get loaded the first time": Previously `%s` was replaced with `(mapconcat #'symbol-name ',modes "/")`.  Now use `buffer-file-name` or `load-file-name`.

But I'm beginning to think langfmt is not the way to go because of `SHOW-ERRORS-VAR`.  langfmt defines `SHOW-ERRORS-VAR` for each formatting tool, but I think it should provide single customizable variable that works for all formatting tools.  I'm concerned that this may make the code more difficult for me.

dominikh:
- We can remove the warning about autoloading ("Note that this
  will cause %s to get loaded the first time") now. It was only
  an issue when people called the function unconditionally. Now
  it should only be called when the mode is active, which means
  it has already been loaded.

- We should remove the check for the major mode from
  *-before-save. People aren't supposed to call it in the wrong
  mode, so we shouldn't check for it, either.

- Regarding SHOW-ERRORS-VAR: It should be trivial to have one
  single variable for this, defined in a langfmt group, used by
  all the functions. What'd be the issue with this?

  Alternatively, the design of langfmt could be slightly
  modified. Instead of being a set of macros, it could provide a
  variable with a mapping of major mode -> formatting functions,
  and a single before-save hook that runs the right function.
  Instead of each major mode defining their own functions via
  macros, they could just append to the langfmt mapping, and the
  user would run a langfmt minor mode.

papaeye:
It doesn't make sense to me.  Could you explain it in more detail?  I haven't changed BEFORE-SAVE-FUNC for backward compatibility.  No need for backward compatibility?

> What'd be the issue with this?

I thought it's strange to expose the langfmt variable to users of formatting functions.  If exposing it, langfmt should be a framework or tool (not a library) as you mentioned.

dominikh:
Backwards compatibility isn't needed in langfmt. I'll overwrite/advise the function myself in go-mode. langfmt, as a new library, shouldn't be burdened with mistakes that we can fix now.

What's left is deciding whether langfmt should be a library/set of macros, or a minor mode that people interact with directly. I'm fine with either, but am leaning towards the latter. A unified framework for formatting code, with settings that pertain to all major modes at once, seems like something that could benefit users. What's your opinion? 

We should turn this PR into a new repository regardless of the decision regarding framework/library, though, before we progress. I do think that langfmt should exist in one way or another. The PR in its current form is a good foundation. If we want to turn it into a minor mode, that should be done in a separate PR on the new repository.

papaeye:
> langfmt, as a new library, shouldn't be burdened with mistakes that we can fix now.

Makes sense.  Replaced BEFORE-SAVE-FUNC with the definition of RUN-BEFORE-SAVE-FUNC and removed RUN-BEFORE-SAVE-FUNC.

> What's your opinion?

I like the latter too.  go-mode.el is the very good foundation.  Thanks for your great work!

dominikh:
I've merged your patch in https://github.com/dominikh/langfmt.el – I also filed an issue in that repository to track planned changes to the API. Future discussions should be held in that repository.

PR #113, opened 2015-12-14
gofmt: Wrap external commands in `unwind-protect'

juergenhoetzel:
This ensures that the temporary file and buffer gets deleted.


dominikh:
Merged as 15d227ae1cf8d55ae4395108bccc8ad6ca7d1eba, thanks.

PR #115, opened 2016-01-04
Update README.md

hus787:
github.com/tw4452852/godef is improved over github.com/rogpeppe/godef. So better recommend the former over the latter.


dominikh:
No thanks. Rog's godef is the canonical one and the one we'll
recommend. What happens if there's 2 more forks, all with their own
improvements?

Furthermore, the improvements in that fork are minimal, and I'm not
even sure if the change regarding listing more than one match works
with go-mode.

Either way, we'll keep recommending the official godef, and I
sincerely hope that any noteworthy improvements will be contributed
upstream. We don't need a fragmented ecosystem.

PR #117, opened 2016-01-09
Add goto functions for working with function signatures

thiderman:
Add several functions:
- `go-goto-arguments`
- `go-goto-docstring`
- `go-goto-function`
- `go-goto-function-name`
- `go-goto-return-value`
- `go-goto-type-signature`

They are all about going to different parts of function signatures, with
their names being self descriptive. They all operate on the current
function, the definition of which is "the closest `func` statement above
point". If no `func` is found above point, the functions will jump to
the first `func` in the file. Lastly, if the point is in the docstring
of a function, the goto functions will operate on that function and not
the function above.go-goto-type-signature

Additionally, some of the functions will create their targets if they are
missing:
- `go-goto-docstring` will create a docstring if it does not
  exist as well as update the docstring name for a function that has been
  renamed.
- `go-goto-return-value` will add space for adding return values.
- `go-goto-type-signature` will add spaces and a pair of parenthesis for
  adding a signature.

It should also be noted that the functions have been tested with
anonymous functions as well.

Add keymap at `C-c g` for goto functions.
Add mapping for the already existing `go-goto-imports` at `C-c g i`.
Add helper function `go--get-function-name`, which grabs the string of
the function above point.

Update README to reflect the addition of these functions.

---

I have been using these functions for about a month now. They feel stable enough that they can be shared with the upstream `go-mode`.

It should be noted and is probably apparent that this is my first real contribution to a project written in elisp (or any lisp at all, actually :dancer:). As such, I am probably missing out on conventions, useful functions and other such things. I'd be very happy to change the code so that it fits well into the project! Comments on coding style and such things are extremely welcome!

---

## TODO

- [x] `go-goto-docstring` should not update the docstring
- [x] Keybinds should not be on the user bindable keyspace.
- [x] Move all `define-key` to one single place.
- [x] `go-goto-arguments` has a copy-pasted docstring
- [x] ~~Handle going to docstrings with prepending words (`A foo` etc)~~
- [x] Fix corner cases for `go-goto-function`
- [x] Don't skip past the pointer in `go-goto-return-value`
- [x] Use `forward-list` instead of searching for `)`.
- [x] `go-goto-type-signature` has the wrong name.
- [x] Docstring and comments for `go--in-function-p` are wrong
- [x] `go-goto-docstring` has a bug with empty comments above a function
- [x] Add prefix arguments and handling of special cases for anonymous functions
- [x] Fix bug in `go-goto-docstring` in the empty comments case
- [x] Remove `get-` from `go--get-function-name`
- [x] Handle prefix arguments as booleans
- [x] Fix logical errors in `go--in-function`
- [x] Handle prefixed non-interactive calls to `go-goto-docstring` and `go-goto-method-receiver`.
- [x] Make `go--function-name` return `nil` for anonymous functions.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -600,7 +600,7 @@ current line will be returned."
     ;; breaks if there's a comment between the struct/interface keyword and
     ;; bracket, like this:
     ;;
-    ;;     struct /* why? */ { 
+    ;;     struct /* why? */ {
     (while (progn
       (skip-chars-forward "^{")
       (forward-char)
--- a/go-mode.el
+++ b/go-mode.el
@@ -1586,11 +1586,12 @@ are below that anonymous function, go to the root function."
         (looking-at "^//"))
       ;; In case we are looking at the docstring, move on forward until we are
       ;; not anymore
-      ;; TODO(thiderman): This would behave incorrectly if point is
-      ;; inside a standalone block of comments that are not a docstring.
       (beginning-of-line)
       (while (looking-at "^//")
-        (forward-line 1)))
+        (forward-line 1))
+      ;; If we are still not looking at a function, retry by calling self again.
+      (when (not (looking-at "^func"))
+        (go-goto-function)))
 
      ((not (looking-at "^func"))
       ;; If we are not looking at the beginning of a function line, do a regexp
@@ -1600,7 +1601,8 @@ are below that anonymous function, go to the root function."
       ;; If nothing is found, assume that we are at the top of the file and
       ;; should search forward instead.
       (when (not (looking-at "func"))
-        (re-search-forward "\\<func\\>" nil t))
+        (re-search-forward "\\<func\\>" nil t)
+        (forward-word -1))
 
       ;; If we have landed at an anonymous function, it is possible that we
       ;; were not inside it but below it. If we were not inside it, we should
--- a/go-mode.el
+++ b/go-mode.el
@@ -1624,8 +1624,9 @@ If the function is anonymous, place point on the 'func' keyword."
 If there is none, make space for one to be added."
   (interactive)
   (go-goto-arguments)
-  (re-search-forward ")" nil t)
-  (forward-char 1)
+  (backward-char)
+  (forward-list)
+  (forward-char)
 
   ;; Opening parenthesis, enter it
   (when (looking-at "(")
--- a/go-mode.el
+++ b/go-mode.el
@@ -1612,7 +1612,7 @@ If the function is anonymous, place point on the 'func' keyword."
         (forward-char 4)))))
 
 (defun go-goto-arguments ()
-  "Go to the return value declaration of the current function."
+  "Go to the arguments of the current function."
   (interactive)
   (go-goto-function-name)
   (forward-word 1)
--- a/go-mode.el
+++ b/go-mode.el
@@ -1712,21 +1712,31 @@ If there is none, add parenthesis to add one."
 (defun go-goto-docstring ()
   "Go to the top of the docstring of the current function.
 
-If there is none, add one."
+If there is none, add slashes to start writing one."
   (interactive)
   (go-goto-function)
   (forward-line -1)
   (beginning-of-line)
+
   (while (looking-at "^//")
     (forward-line -1))
-  (next-line 1)
+  (forward-line 1)
   (beginning-of-line)
-  (if (not (looking-at "^func"))
-      (forward-char 3)
-    ;; If we are still at the function signature, we should add a new docstring.
+
+  (cond
+   ;; If we are looking at an empty comment, add a single space in front of it.
+   ((looking-at "^//$")
+    (forward-char 2)
+    (insert " "))
+   ;; If we are not looking at the function signature, we are looking at a docstring.
+   ;; Move to the beginning of the first word of it.
+   ((not (looking-at "^func"))
+    (forward-char 3))
+   ;; If we are still at the function signature, we should add a new docstring.
+   (t
     (forward-line -1)
     (newline)
-    (insert (format "// %s " (go--get-function-name)))))
+    (insert (format "// %s " (go--get-function-name))))))
 
 (defun go--get-function-name ()
   "Return the current function name as a string"
--- a/go-mode.el
+++ b/go-mode.el
@@ -1636,8 +1636,8 @@ If there is none, make space for one to be added."
     (insert " ")
     (backward-char 1)))
 
-(defun go-goto-type-signature ()
-  "Go to the type signature of the current function.
+(defun go-goto-method-receiver ()
+  "Go to the receiver of the current method.
 
 If there is none, add parenthesis to add one."
   (interactive)
@@ -1680,8 +1680,8 @@ If there is none, add one."
 (define-key go-goto-map (kbd "d") 'go-goto-docstring)
 (define-key go-goto-map (kbd "f") 'go-goto-function-name)
 (define-key go-goto-map (kbd "i") 'go-goto-imports)
+(define-key go-goto-map (kbd "m") 'go-goto-method-receiver)
 (define-key go-goto-map (kbd "r") 'go-goto-return-value)
-(define-key go-goto-map (kbd "t") 'go-goto-type-signature)
 
 (provide 'go-mode)
 
--- a/go-mode.el
+++ b/go-mode.el
@@ -1757,7 +1757,7 @@ an error is raised."
    ;; If we are looking at an empty comment, add a single space in front of it.
    ((looking-at "^//$")
     (forward-char 2)
-    (insert (format " %s " (go--function-name))))
+    (insert (format " %s " (go--function-name t))))
    ;; If we are not looking at the function signature, we are looking at a docstring.
    ;; Move to the beginning of the first word of it.
    ((not (looking-at "^func"))
@@ -1766,15 +1766,20 @@ an error is raised."
    (t
     (forward-line -1)
     (newline)
-    (insert (format "// %s " (go--function-name))))))
+    (insert (format "// %s " (go--function-name t))))))
 
-(defun go--function-name ()
+(defun go--function-name (&optional arg)
   "Return the current function name as a string.
 
-Will skip anonymous functions since they do not have names."
-  (save-excursion
-    (go-goto-function-name t)
-    (symbol-name (symbol-at-point))))
+If `arg' is non-nil anonymous functions will be ignored and the
+name returned will be of the top-level function.
+
+Returns nil otherwise."
+  (when (or (not (go--in-anonymous-funcion-p))
+            arg)
+    (save-excursion
+      (go-goto-function-name t)
+      (symbol-name (symbol-at-point)))))
 
 (defun go--in-anonymous-funcion-p ()
   "Return t if point is inside an anonymous function, nil otherwise."
--- a/go-mode.el
+++ b/go-mode.el
@@ -1569,17 +1569,16 @@ for."
           (display-buffer (current-buffer) `(,go-coverage-display-buffer-func))))))
 
 (defun go-goto-function (&optional arg)
-  "Go to the function defintion above point.
+  "Go to the function defintion (named or anonymous) surrounding point.
 
 If we are on a docstring, follow the docstring down.
 If no function is found, assume that we are at the top of a file
 and search forward instead.
 
-If we are inside an anonymous function, go to that.
-If we are inside a function that has an anonymous function inside of it and we
-are below that anonymous function, go to the root function.
+If point is looking at the func keyword of an anonymous function,
+go to the surrounding function.
 
-If one prefix argument is given, anonymous functions are skipped."
+If ARG is non-nil, anonymous functions are ignored."
   (interactive "P")
   (let ((p (point)))
     (cond
@@ -1592,13 +1591,22 @@ If one prefix argument is given, anonymous functions are skipped."
       (while (looking-at "^//")
         (forward-line 1))
       ;; If we are still not looking at a function, retry by calling self again.
-      (when (not (looking-at "^func"))
+      (when (not (looking-at "func"))
         (go-goto-function arg)))
 
-     ((not (looking-at "^func"))
-      ;; If we are not looking at the beginning of a function line, do a regexp
-      ;; search backwards
-      (re-search-backward "\\<func\\>" nil t)
+     ;; If we're already looking at an anonymous func, look for the
+     ;; surrounding function.
+     ((and (looking-at "func")
+           (not (looking-at "^func")))
+      (re-search-backward "\\<func\\>" nil t))
+
+     ((not (looking-at "func"))
+      ;; If point is on the "func" keyword, step back a word and retry
+      (if (string= (symbol-name (symbol-at-point)) "func")
+          (backward-word)
+        ;; If we are not looking at the beginning of a function line, do a regexp
+        ;; search backwards
+        (re-search-backward "\\<func\\>" nil t))
 
       ;; If nothing is found, assume that we are at the top of the file and
       ;; should search forward instead.
@@ -1609,7 +1617,7 @@ If one prefix argument is given, anonymous functions are skipped."
       ;; If we have landed at an anonymous function, it is possible that we
       ;; were not inside it but below it. If we were not inside it, we should
       ;; go to the containing function.
-      (while (go--below-anonymous-function p)
+      (while (not (go--in-function-p p))
         (go-goto-function arg)))))
 
   (cond
@@ -1622,32 +1630,30 @@ If one prefix argument is given, anonymous functions are skipped."
     ;; been supplied, redo the call so that we skip the anonymous function.
     (go-goto-function arg))))
 
-(defun go--below-anonymous-function (compare-point)
-  "Return t if `compare-point' is below the function that point is currently on.
-This should only be called as a helper when point is looking at \"func\".
+(defun go--goto-opening-curly-brace ()
+  (go--goto-return-values)
+  ;; Try to place the point on the opening brace.
+  (cond
+   ((looking-at "(")
+    (forward-list 1)
+    (forward-char 1))
+   ((not (looking-at "{"))
+    (forward-word 1)
+    (forward-char 1))))
 
-Returns nil in all other cases."
-  ;; Check that we are not looking at a top level function. If we are, return nil.
-  (when (not (looking-at "^func"))
-    (save-excursion
-      (go--goto-return-values)
-      ;; Try to place the point on the opening brace.
-      (cond
-       ((looking-at "(")
-        (forward-list 1)
-        (forward-char 1))
-       ((not (looking-at "{"))
-        (forward-word 1)
-        (forward-char 1)))
+(defun go--in-function-p (compare-point)
+  "Return t if COMPARE-POINT lies inside the function immediately surrounding point."
+  (save-excursion
+    (when (not (looking-at "func"))
+      (go-goto-function))
+    (let ((start (point)))
+      (go--goto-opening-curly-brace)
 
-      (when (looking-at "{")
-        ;; Go past the body of the function and back inside of it.
-        (forward-list 1)
-        (backward-char 1)
-        ;; Now that point is looking at the closing brace of the function
-        ;; body, compare if our position is earlier in the file than the
-        ;; comparing point. Return t if it is.
-        (< (point) compare-point)))))
+      (unless (looking-at "{")
+        (error "expected to be looking at opening curly brace"))
+      (forward-list 1)
+      (and (>= compare-point start)
+           (<= compare-point (point))))))
 
 (defun go-goto-function-name (&optional arg)
   "Go to the name of the current function.
--- a/go-mode.el
+++ b/go-mode.el
@@ -1630,9 +1630,6 @@ If there is none, make space for one to be added."
   ;; Opening parenthesis, enter it
   (when (looking-at "(")
     (forward-char 1))
-  ;; First item is a pointer, move past the pointer
-  (when (looking-at "*")
-    (forward-char 1))
   ;; No return arguments, add space for adding
   (when (looking-at "{")
     (insert " ")
--- a/go-mode.el
+++ b/go-mode.el
@@ -1617,7 +1617,8 @@ If ARG is non-nil, anonymous functions are ignored."
       ;; If we have landed at an anonymous function, it is possible that we
       ;; were not inside it but below it. If we were not inside it, we should
       ;; go to the containing function.
-      (while (not (go--in-function-p p))
+      (while (and (not (go--in-function-p p))
+                  (not (looking-at "^func")))
         (go-goto-function arg)))))
 
   (cond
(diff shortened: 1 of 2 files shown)


dominikh:
Before I actually look at the code, and test corner cases, here's my input on the proposed functions:

- `go-goto-function`: How is it different from `beginning-of-defun`? Is the only difference that it works with anonymous functions?
- `go-goto-docstring`: I don't think it should update an existing docstring. For one, it's not implied by the name. For another, there are probably valid cases where the docstring doesn't match the function name in an expected way and shouldn't be changed

About the key bindings: Those are not okay. `C-c <letter>` is reserved for users. See http://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Binding-Conventions.html – It's why the existing bindings all are `C-c C-<letter>`

go-mode.el
```diff
@@ -1562,6 +1568,143 @@ for."
...
+          (chars 1))
+      (when (looking-at "func (")
+        (setq words 3
+              chars 2))
+      (forward-word words)
+      (forward-char chars)
+      (when (looking-at "Test")
+        (forward-char 4)))))
+
+(define-key go-goto-map (kbd "f") 'go-goto-function-name)
```
dominikh: Please move all `define-key` to a single place. Optimally below the definition of go-mode-map.

go-mode.el
```diff
@@ -1562,6 +1568,143 @@ for."
...
+
+(defun go-goto-arguments ()
+  "Go to the return value declaration of the current function."
```
dominikh: Wrong docstring

thiderman:
* `go-goto-function` also handles the case where you are on the docstring of a function. `beginning-of-defun` will go to the one above regardless. Do you think a better path would be to extend `beginning-of-defun`?
* For `go-goto-docstring` I can remove that part from the goto function and add a separate function that can update the name!

Regarding keybindings, I did not know that was a thing, and now I do! Thanks! I'll fix it. The document doesn't really specify, but I think that `C-c C-g f` etc would suffice, right? `C-c C-g C-f` just seems extreme.

Thanks for the quick feedback! :+1:

dominikh:
- I don't think we can change the behaviour of beginning-of-defun – Right now, repeatedly calling it will move upwards through all functions. If it behaved like `go-goto-function`, it would get stuck on the first function with a docstring. I *suppose* it is fine to have both functions.
- `C-c C-g f` would suffice, yes.

dominikh:
And yes, please remove that part of `go-goto-docstring`. Also, when you extract that functionality into its own function, fix the following bugs:

- "A foo" and "The foo" are valid alternatives to "foo"
- An empty docstring (`//`) currently deletes the `func` keyword and replaces it with the function name
- A docstring consisting of just a space (`// `) does the same as `//`, but additionally pulls a line of code into the docstring

dominikh:
`go-goto-function` fails for the following piece of code, with point on `"hi"`:

```
package main

func the_func() {
	_ = func() { println("hi") }
}
```

dominikh:
`go-goto-function` fails for the following piece of code, with point on `"bye"`

```
package main

func the_func() {
	_ = func() {
		// func() {
		//   println("hi")
		// }
		println("bye")
	}
}
```

Take a look at `go-beginning-of-defun`, `go-in-string-or-comment-p` and `go--backward-irrelevant`

dominikh:
Interim: I do like the additions in general, but be aware that this functionality will have to deal with a lot of weird corner cases, so don't be too surprised when I throw a lot of broken examples at you.

thiderman:
> Interim: I do like the additions in general, but be aware that this functionality will have to deal with a lot of weird corner cases, so don't be too surprised when I throw a lot of broken examples at you.

I expected nothing else! :smile: I've been doing some plugin development for vim, so I am used to the craziness of corner cases. However, I have since seen the light and am now enjoying developing editor things in an environment that actually wants you to edit it, haha.

Thanks for the pointers about the functions! I'll see if I can make them work for the test cases!

dominikh:
- In `go-goto-return-value`, please don't skip past the pointer. That'll just trip people up.
- `go-goto-return-value` fails for the following code:

```
func foo(i interface {
	Foo()
}) int {
	return 0
}
```

You should use `forward-list` instead of searching for the first ")" you can find.

dominikh:
`go-goto-type-signature` is not named correctly. It goes to the method's receiver. The type signature of a function is something else (The type signature of `func foo(a int) (b int)` is something along the lines of `func (int) (int))`

thiderman:
Do you prefer me just amending the commits, or do you want separate commits per fix?

> go-goto-type-signature is not named correctly. It goes to the method's receiver. The type signature of a function is something else (The type signature of func foo(a int) (b int) is something along the lines of func (int) (int))

Noted and will be fixed!

dominikh:
`go-goto-type-signature` does the wrong thing for the following code, with point on `"hi"`:

```
package main

func foo() {
	_ = func() {
		println("hi")
	}
}
```

dominikh:
Separate commits per fix please. We'll squash them into one commit at the end, but during the review process, multiple commits help tremendously with keeping track of outstanding issues.

thiderman:
Trying to use `forward-list`, I run into errors:

```
Debugger entered--Lisp error: (scan-error "Containing expression ends prematurely" 3420 3421)
  scan-lists(3420 1 0)
  forward-list(1)
  call-interactively(forward-list nil nil)
  command-execute(forward-list)
```

This seems to happen whenever this is executed from inside of a pair of parenthesis. This seems to be true everywhere in the go files, not just for the case of empty arguments (as I happened to be testing). Do you know what could be the cause of this?

dominikh:
You want to move point before the opening `(`

thiderman:
Ah, of course. I'll fix that!

Also, IRL calls and I'll have to leave for the night. I'll pick up tomorrow! Thanks so much for the help and the feedback! :smile:

thiderman:
> "A foo" and "The foo" are valid alternatives to "foo"

According to the [official blog post](http://blog.golang.org/godoc-documenting-go-code) about docstrings, the convention is to always start with the element the docstring describes. Furthermore [golint](https://github.com/golang/lint) complains with something like `tournament.go:146:1: comment on exported method Tournament.Start should be of the form "Start ..."`

Should I still edit the docstring functions to consider those cases?

dominikh:
No, disregard me. The rule that I mentioned, which allows articles, only applies to types, not functions.

thiderman:
Aight, got it! When/if I add support for adding docstrings to types I'll remember that!

I'm now struggling with what to call `go-goto-type-signature`. go-goto-receiver` is fine, but I don't know what key to put it on. `C-c C-g r` is already taken by `go-goto-return-value`. 

Maybe `C-c C-g v` for `go-goto-return-value` and `C-c C-g r` for this one?
What do you think?

dominikh:
How about `C-c C-g m` for `go-goto-method-receiver`?

thiderman:
I like it!

thiderman:
Now I'm just going to tackle all of the edge cases. This will take longer, heh.

Does it makes sense to add a file or two to into a `goto_tests/` directory so that cases are in the repo?

dominikh:
I think I'll deal with goto_tests myself. In particular, I finally/one day want to automate them. I've tried more edge cases than I mentioned here (because some worked fine.)

dominikh:
Another issue: go-goto-function fails for the following code, with point on `"bye"`. It jumps to the wrong function.

```
package main

func main() {
	_ = func() {
		println("hi")
	}

	println("bye")
}
```

thiderman:
Sounds good. I'll just work to fix the already mentioned cases then!

Oh, that last one is tricky indeed. Makes me think that a simple module with some [ast](https://golang.org/pkg/go/ast/) trickery might be the best way to handle all of the special cases. As far as I can figure, the only way of correctly detecting that case and not going to the anonymous function would be by parsing it completely. That would actually solve most of the corner cases, I figure.

I'd love to work on something like that (I can imagine lots of other cool things this major mode could do if it could parse the code fully!), but maybe as a second iteration of this work. What do you think?

dominikh:
Well, `go/ast` isn't an option. I do not want to add additional external dependencies, especially not for core features. So calling a binary for this is a no-go. An AST itself, computed via elisp, would be an option, but also add a significant amount of complexity to go-mode. I want to avoid that, too. (Same goes for relying on CEDET/Semantic or the sorts.)

The particular corner case that I posted has a simpler fix: 

1. remember `point`
2. go to the first (innermost) function you find
3. go to the end of that function (`for-ward-list` on its opening curly brace) 
4. check if the remembered point lies within the function you just found. If not, repeat 2-4 with the next function.

The most tricky bit is finding the opening curly brace. You will again want to use `forward-list` and not get confused by curly braces in types in the arguments or return values. You also want to work correctly in case the function has no return values.

thiderman:
I see, and that makes sense. I'll see if I can figure out some other nice features that won't require anything but elisp! Pull Requests and Issues might pop up in the future! :)

Oh, that's a pretty neat way. I'll tinker with that!
Is curly braces in arguments and return values really a thing? When? I've never seen anything but alphanumerics, because the expressions are always pointing at a type.

dominikh:
One of my examples only demonstrates one such case: anonymous interface types. While these are relatively uncommon, there are also anonymous struct types, which are more common.

thiderman:
Ah, of course. That case is mostly easy to handle though since they always come in immediately closing pairs like `{}`, right?

dominikh:
Immeditately closing pairs? Anonymous types can look like any other type definition, minus the name. That's really not relevant, though. They are balanced pairs of parantheses, and `forward-list` will do the right thing if you're on the opening parenthesis of the argument list or return value list.

thiderman:
There! Took a bit longer than intended, but I think I have fixed all the corner cases.

`go-goto-function` is now aware of anonymous functions and handles the case if we are inside an anonymous function. _Inside_ is defined as "somewhere inside the body wrapped by braces". 
Since all the other `goto` functions added here use `go-goto-function` somehow, this adds support for all of them.

I also made sure that nested anonymous functions are supported as well:
```
func wrapping() {
        _ = func() {
                _ = func() {
                        println("hi")
                }
        }
        println("bye")
}
```

I think this solves all the errors you found! :) Did I miss something?

The only thing I have left to figure out what to do when `go-goto-method-receiver` is called inside an anonymous function. I don't think an anonymous function can have a receiver, so I don't know what `go-goto-method-receiver` should do then. Not move at all with an error message? What do you think?

dominikh:
> Did I miss something or add new errors?

I discovered some new errors.

For the following code, `go-goto-docstring` doesn't work correctly:

```
package main

//
func foo() {

}
```

Note that there is no trailing space after the `//` in my example.

For the following code, with point on `"hi"`, `go-goto-docstring` does the wrong thing:

```
package main

func main() {
    _ = func() {
        println("hi")
    }

    println("bye")
}
```

But see further down for some additional information on that.

> Not move at all with an error message? What do you think

I'm torn. When calling `go-goto-method-receiver` programmatically, we probably want the error. However, when it's called interactively, the user most likely expects it to jump to the method's receiver, and ignore anonymous functions. The same could be said for `go-goto-docstring`.

This issue also leads to another question: Should we add a prefix argument (`C-u`) to always ignore anonymous functions for all of the movement commands? That is, allow a user to use `C-u go-goto-arguments` to jump to the named function's arguments, instead of those of the anonymous function point is in? If we went for that idea, all functions would obtain an optional argument, which would default to nil for all functions but `go-goto-docstring` and `go-goto-method-receiver`. The latter two would default to nil if called non-interactively, or `(4)` if called interactively. If the value is nil, and point is in an anonymous function, an error would be thrown. What do you think?


I haven't yet reviewed the new code, I'll get back to you on that.

dominikh:
Also I've changed my mind about squashing the commits. I think we should keep all of the commits intact, They're likely to aid in debugging issues in the future.

go-mode.el
```diff
@@ -1562,6 +1568,185 @@ for."
...
+       (forward-char 1))
+      ((not (looking-at "{"))
+       (forward-word 1)
+       (forward-char 1)))
+
+     (when (looking-at "{")
+       ;; Go past the body of the function and back inside of it.
+       (forward-list 1)
+       (backward-char 1)
+       ;; Finally, compare if our position is past the assert point.
```
dominikh: This comment confuses me. Your code checks if `point < compare-point` – how is that being "past the assert point"?

go-mode.el
```diff
@@ -1562,6 +1568,185 @@ for."
...
+      ;; If nothing is found, assume that we are at the top of the file and
+      ;; should search forward instead.
+      (when (not (looking-at "func"))
+        (re-search-forward "\\<func\\>" nil t)
+        (forward-word -1))
+
+      ;; If we have landed at an anonymous function, it is possible that we
+      ;; were not inside it but below it. If we were not inside it, we should
+      ;; go to the containing function.
+      (while (go--in-function-p p)
```
dominikh: I'm not sure I understand why this works. If the original point was in the function that we arrived at, shouldn't we stop, instead of continuing? Testing shows that the code does in fact work. But why?

go-mode.el
```diff
@@ -1562,6 +1568,185 @@ for."
...
+        (go-goto-function)))))
+
+  ;; If we are still in a comment, redo the call so that we get out of it.
+  (when (go-in-comment-p)
+    (go-goto-function)))
+
+(defun go--in-function-p (compare-point)
+  "Return t if point is inside the function that starts at `compare-point', nil
```
dominikh: Is this docstring accurate? Does the function *start* at `compare-point`, or is `compare-point` just an arbtirary point inside the function? The way it's being called from `go-goto-function`, I would've assumed the latter.

thiderman: It's not. It was rewritten (first it got `point` as an argument, but then I removed that and just went with where point actually is) and I missed that. I'll fix it!

go-mode.el
thiderman: Internally that function will call `go-goto-function` in every iteration. If point is at `<point>func`, `re-search-backwards` will search backwards further, so for every time `go--in-function-p` is `t` the loop will continue upwards until the end of the cursor is past the comparing point `p`.
The first check in `go--in-function-p` (`(looking-at "^func")`) will make it stop if it reaches a top level function (which I think only happens when you are outside a top-level function).

It's basically the algorithm you described, but wrapped in a `while` loop.

I'd agree it is a tad complicated, but I couldn't figure out a simpler way of doing it. Do you want me to try to add a better explanation in the comments?

go-mode.el
thiderman: Wrong in the same vein as the docstring. I'll fix it!

thiderman:
I like the idea of using prefix arguments and defaulting the other ones! I'll tinker with that and see what I can do! :dancer:

go-mode.el
dominikh: It feels to me like `go--in-function-p` returns whether `compare-point` is *NOT* in the function that point is at. Consider this piece of code:

```
package main

func main() {
	_ = func() {
		//
	}
	_ = 1
}
```

With `compare-point` on `1` and point looking at the anonymous `func`, `go--in-function-p` returns t. With `point` looking at the named func, it returns nil. Shouldn't the opposite be the case? `compare-point` isn't in the anonymous function, but it is in the named function. 

That's why the while loop feels confusing to me, too. If `go--in-function-p` worked as I expected, the while loop would have to check for `(not (go--in-function-p))`.

What am I missing?

thiderman: You are completely right. I've made a blunder with `<` vs `>` and then just considered it working as soon as the case was okay. This is a case of two wrongs kind of making a right.
I'll refactor that right away.

thiderman:
I have addressed the latest batch of feedback.

The only thing I didn't fix was the naming issue of `go--in-function-p`. I tried flipping the logic around so that the calling loop was `(while (not (go--in-function-p)))` and flipping the logic inside of the function itself. But, it always turned out wrong in some of the corner cases, mostly because the loop didn't want to exit. Could be that I am not smart enough to fix it, or that I am missing something simple.

Anyway, as you already pointed out the name is kind of off anyway. The function doesn't test for if it is inside anything, it tests if point is *below an anonymous function*, just as you described in the algorithm. How about just renaming the current implementation to something else for clarity? I was thinking `go--below-anonymous-function`. What do you think about that?


Also, huge thanks for bearing with me and my noobish pull request! This has been (and still is) fun and rewarding for me, and your feedback is awesome! This is easily the best open source experience I have had in... ever, and you deserve recognition and thanks for that! Cheers!

dominikh:
> What do you think about that?

I'll take my own crack at it and see if I can get a proper `go--in-function-p` working. I think such a function could be useful in the future. If that fails we can still think about a new name.

> Also, huge thanks for bearing with me and my noobish pull request

Thanks for bearing with me and fixing all the issues, instead of abandoning the PR :-) It's been a nice experience for me as well. And the PR was by no means noobish. Your code mostly worked, was well written and only needed some fixes. And you've made correct fixes given my input.

dominikh:
@thiderman can you confirm that the following patch passes all your tests?

```
From 14ab5c29c44a562dccc409737a8fbe4a6a34a42d Mon Sep 17 00:00:00 2001
From: Dominik Honnef <dominik@honnef.co>
Date: Wed, 20 Jan 2016 03:55:43 +0100
Subject: [PATCH] Reimplement go--in-function-p to do as the name says

---
 go-mode.el | 48 ++++++++++++++++++++++--------------------------
 1 file changed, 22 insertions(+), 26 deletions(-)

diff --git a/go-mode.el b/go-mode.el
index 0da9e13..0c4de38 100644
--- a/go-mode.el
+++ b/go-mode.el
@@ -1609,7 +1609,7 @@ If one prefix argument is given, anonymous functions are skipped."
       ;; If we have landed at an anonymous function, it is possible that we
       ;; were not inside it but below it. If we were not inside it, we should
       ;; go to the containing function.
-      (while (go--below-anonymous-function p)
+      (while (not (go--in-function-p p))
         (go-goto-function arg)))))
 
   (cond
@@ -1622,32 +1622,28 @@ If one prefix argument is given, anonymous functions are skipped."
     ;; been supplied, redo the call so that we skip the anonymous function.
     (go-goto-function arg))))
 
-(defun go--below-anonymous-function (compare-point)
-  "Return t if `compare-point' is below the function that point is currently on.
-This should only be called as a helper when point is looking at \"func\".
+(defun go--goto-opening-curly-brace ()
+  (go--goto-return-values)
+  ;; Try to place the point on the opening brace.
+  (cond
+   ((looking-at "(")
+    (forward-list 1)
+    (forward-char 1))
+   ((not (looking-at "{"))
+    (forward-word 1)
+    (forward-char 1))))
 
-Returns nil in all other cases."
-  ;; Check that we are not looking at a top level function. If we are, return nil.
-  (when (not (looking-at "^func"))
-    (save-excursion
-      (go--goto-return-values)
-      ;; Try to place the point on the opening brace.
-      (cond
-       ((looking-at "(")
-        (forward-list 1)
-        (forward-char 1))
-       ((not (looking-at "{"))
-        (forward-word 1)
-        (forward-char 1)))
-
-      (when (looking-at "{")
-        ;; Go past the body of the function and back inside of it.
-        (forward-list 1)
-        (backward-char 1)
-        ;; Now that point is looking at the closing brace of the function
-        ;; body, compare if our position is earlier in the file than the
-        ;; comparing point. Return t if it is.
-        (< (point) compare-point)))))
+(defun go--in-function-p (compare-point)
+  "Return t if COMPARE-POINT lies inside the function immediately surrounding point."
+  (save-excursion
+    (go--goto-opening-curly-brace)
+
+    (unless (looking-at "{")
+      (error "expected to be looking at opening curly brace"))
+    (let ((start (point)))
+      (forward-list 1)
+      (and (>= compare-point start)
+           (<= compare-point (point))))))
 
 (defun go-goto-function-name (&optional arg)
   "Go to the name of the current function.
-- 
2.7.0


```

dominikh:
Also you've failed to address https://github.com/thiderman/go-mode.el/commit/79c5a02d53f18687cb9122198fdaa65977ca204d#commitcomment-15519234

dominikh:
> Furthermore, your examples are violating the conventions: the_func should be written as theFunc. Considering this, the case where there is a false hit because \\<func\\> matches the_func might not be something go-goto-function should worry about. What do you think?

Sorry, I only now saw your question. I violate conventions in examples/test cases on purpose, because in general, go-mode should work fine with code that isn't idiomatic, or even code that isn't gofmt'ed correctly. That is, go-mode should work with any Go code that follows the spec. For example, whitespace that is optional in the spec should also be optional in go-mode. And any valid identifier should work, too.

There are some rare cases where we do require gofmt'ed code, because the alternative would be too complex to handle. None of the functions you've added, however, qualify for that.

I do see, however, that the `the_func` case is being handled correctly now. I'll check for (and possibly fix myself) any other assumptions during my final code review.

dominikh:
Okay, I've found at least one corner case that fails with my patch. I'll work on it some more.

thiderman:
> And the PR was by no means noobish.

Thanks for the kind words! :smile: 

> Also you've failed to address thiderman@79c5a02#commitcomment-15519234

Ah, right - my bad. Easily fixed!

> I do see, however, that the the_func case is being handled correctly now.

Yeah, back when I asked that that function used word boundaries to find functions. It has since stopped doing that and it works fine. I think you can disregard that question. :+1: 

> Okay, I've found at least one corner case that fails with my patch. I'll work on it some more.

Sounds good! I'll wait meanwhile! 
I also added you as a collaborator to my fork, so you can push directly to this branch and you won't need to do the mail-style patch thing. Easier for both of us, I think. :8ball:

dominikh:
I've fixed `go--in-function-p` and improved `go-goto-function` (plus some other minor fixes). Please check if you can find any new bugs.

Also, I've discovered another issue with your approach. For the following snippet, `go-goto-function` fails:

```
func foo() a.B {
	// hi
}
```

The reason being that `forward-word` will only skip over `a`, not `a.B`.

thiderman:
Both of your commits look good, especially the second one. I'm not used to the emacs idioms yet, heh.

I'll look at making a fix for that case!

thiderman:
As I fixed the case for the anonymous interface, it also fixed the `a.B` case. Yay!

Now I'll go through testing all of the changes to see if everything is still working nicely!

dominikh:
`go--goto-opening-curly-brace` fails for this snippet:

```
func fooInterfaceReturn() map[struct{}]int {
	return nil
}
```

as well as

```
func fooInterfaceReturn() struct{} /* how bad could this be? */ {
	return struct{}{}
}
```

dominikh:
I was going to suggest using `go--at-function-definition`, but unfortunately that suffers the same issues.

dominikh:
However, you should be able to combine `end-of-defun` and `backward-list` to find the opening curly brace.

dominikh:
@thiderman I'm happy to merge this as is, given you didn't find any new bugs. The issue with inline block comments isn't all that important to fix, other go-mode functions haven't supported those in ages and nobody really complained.

I'm sure the new functions can further be tweaked, but we can do that incrementally as new issues show up.

dominikh:
(I'll revert 702080a99278e64e208daedb613d2bf01a5a204b and reimplement it myself, because my suggestion of using end-of-defun was ill-advised. It will break with anonymous functions)

thiderman:
Sounds very good to me! I couldn't find any further problems either!

Did you reimplement the function? If not, do you want me to take a swing at it? :)

PR #120, opened 2016-02-06
Use the go list command to gather list of packages

juergenhoetzel:
Hi Dominik,

I wasn't able to import the "unsafe" package.

Besides simplifying  the implementation of go-packages, this also
allows the import the special "unsafe" package.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -241,6 +241,13 @@ See `display-buffer' for a list of possible functions."
   :type 'function
   :group 'go-cover)
 
+(defcustom go-packages-function 'go-packages
+  "The function called to determine the list of available packages.
+
+Set to `go-packages-external' if you use Cygwin or Tramp filenames."
+  :type 'function
+  :group 'go)
+
 (defface go-coverage-untracked
   '((t (:foreground "#505050")))
   "Coverage color of untracked code."
@@ -1070,7 +1077,7 @@ you save any file, kind of defeating the point of autoloading."
     (completing-read (if symbol
                          (format "godoc (default %s): " symbol)
                        "godoc: ")
-                     (go--old-completion-list-style (go-packages)) nil nil nil 'go-godoc-history symbol)))
+                     (go--old-completion-list-style (funcall go-packages-function)) nil nil nil 'go-godoc-history symbol)))
 
 (defun godoc--get-buffer (query)
   "Get an empty buffer for a godoc query."
@@ -1248,7 +1255,7 @@ uncommented, otherwise a new import will be added."
   (interactive
    (list
     current-prefix-arg
-    (replace-regexp-in-string "^[\"']\\|[\"']$" "" (completing-read "Package: " (go--old-completion-list-style (go-packages))))))
+    (replace-regexp-in-string "^[\"']\\|[\"']$" "" (completing-read "Package: " (go--old-completion-list-style (funcall go-packages-function))))))
   (save-excursion
     (let (as line import-start)
       (if arg
@@ -1321,6 +1328,11 @@ If IGNORE-CASE is non-nil, the comparison is case-insensitive."
                      (go--directory-dirs pkgdir)))))
      (go-root-and-paths)))
    #'string<))
+
+(defun go-packages-external ()
+  (with-temp-buffer
+    (process-file  go-command nil (current-buffer) nil "list" "-e" "...")
+    (split-string (buffer-string) "\n" t)))
 
 (defun go-unused-imports-lines ()
   ;; FIXME Technically, -o /dev/null fails in quite some cases (on


syohex:
`go list` is too slow. It is 10 times slower than original code.

```lisp
(benchmark 1 '(go-packages))
```

Result is here

```
Original: Elapsed time: 0.060883s (0.008502s in 1 GCs)
go list : Elapsed time: 0.676046s
```

Environment
- go version go1.6rc2 linux/amd64
- Ubuntu 15.10
- Intel(R) Core(TM) i7-2620M CPU @ 2.70GHz
- Memory 8GB

dominikh:
I agree that the current solution has its issues, but syohex is point on for why we aren't using `go list`: it is far too slow. Our current solution has far less overhead, and once it's been run once, it caches a lot better.

At the moment, I'm more inclined to address your visible issue, the lack of "unsafe" (and "C"), by adding these two items to the list manually.

juergenhoetzel:
I see the new implementation is slower than the original. But for me its not **too** slow.  To be honest I didn't even notice its slower, because it is an interactive command and the completion list is most likely finished before entering the first character.

New Implementation:
```
ELISP> (benchmark 1 '(go-packages))
"Elapsed time: 0.363932s"
```
Old Implementation:
```
ELISP> (benchmark 1 '(go-packages))
"Elapsed time: 0.064390s"
```

Using a medium size project  (prometheus) on on old Celeron:

Intel(R) Celeron(R) CPU G530 @ 2.40GHz

I would prefer a clean implementation using the "go" tools to an Elisp implementation (depending on the go compiler internals) when speed is not a real issue. Beside the [Cygwin](https://github.com/dominikh/go-mode.el/issues/121) problem, I expect that there are other issues when using the Elisp implementation (vendor directories, error handling).

dominikh:
To demonstrate the underlying performance issues, both for cold and hot file system caches:

```
[cold cache] $ /usr/bin/time -v find $GOPATH/pkg /usr/lib/go/pkg -name "*.a" >/dev/null
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:03.48
[hot cache] $ /usr/bin/time -v find $GOPATH/pkg /usr/lib/go/pkg -name "*.a" >/dev/null
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.01

[cold cache] $ /usr/bin/time -v go list -e ... >/dev/null
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:56.89
[hot cache] $ /usr/bin/time -v go list -e ... >/dev/null
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.39
```

The performance of the current elisp implementation is comparable to that of find. The hot cache performance of `go list` is already too slow for auto-completion, but the real issue is that of the cold-cache performance. Having to wait almost a minute the first time `go-packages` is called is not acceptable.

juergenhoetzel:
Wow, how did you get these numbers? You cleared the filesystem cache by using```echo 3 > /proc/sys/vm/drop_caches ```?

I never got this performance on a real running system. I suspect the user has other performance issues when he clears his cache for any reason :cry:  I suspect ```go build```. has the same performance characteristics in this case.

BTW: Another neat feature would be Tramp support. But this also depends on this commit. I dont think an Elisp implementation would be feasible for this feature, because Tramp would have to traverse the remote paths (quite slow).

@dominikh

Maybe we could introduce a new custom variable like ```go-mode-use-external-go-list```, which is turned of by default and turned on in Tramp directories?

dominikh:
> Wow, how did you get these numbers? You cleared the filesystem cache by using `echo 3 > /proc/sys/vm/drop_caches?`

Yes.

> I suspect the user has other performance issues when he clears his cache for any reason

Anybody who just rebooted his machine will not have his GOPATH cached yet. Anybody who works on machines with relatively little memory may have GOPATH purged from the cache regularly, especially if they're working with other large media.

> I suspect go build. has the same performance characteristics in this case

Only if you manage to build something that depends on every single package in your GOPATH.

I'll admit that my storage isn't the fastest. It's a RAID 5 of 5x5400 RPM disks, so its latency is quite high, compared to that of, say, an SSD. `go list` only has to traverse about 4000 directories in my GOPATH. However, at the same time, I don't want to design go-mode for people with fast machines only, and other people may have significantly larger GOPATHs than me.

> Maybe we could introduce a new custom variable like go-mode-use-external-go-list, which is turned of by default and turned on in Tramp directories?

I don't like the name of that variable, but we can paint that shed later. I'm fine with the idea in general, though. The variable should probably specify a function that returns a list of packages. We'd have two implementations – the current one, and one based on `go list` – and people could write additional ones if they need to, for whatever reason (maybe they use a custom build system that has a different way of iterating packages.)

juergenhoetzel:
How about:

```elisp
(defcustom go-packages-function 'go-packages
  "The function called to determine the list of available packages.

Set to `go-packages-external' if you use Cygwin or Tramp filenames."
  :type 'function
  :group 'go)
```

dominikh:
I'll implement this change next week.

PR #130, opened 2016-03-24
Fix filename replacement in gofmt error buffer

wg:
Hi, thanks for go-mode, it's a critical part of my dev tooling! goimport's new `-srcdir` option rewrites the temp directory path with the original source directory but leaves the temporary `gofmtXXXX.go` filename which breaks regex replacement in `gofmt--process-errors`. Here's a patch to fix that.


dominikh:
Thank you.

PR #133, opened 2016-04-02
Update URL for gogetdoc

zmb3:
Also reccomend use of the -u flag for `go get` since
gogetdoc requires a reasonably up-to-date version of
golang.org/x/tools.


dominikh:
Well, that's an awful copy & paste error, sorry for that.

PR #137, opened 2016-05-12
Dynamically create GOPATH for gofmt-command

enzochiau:
* Dynamically create GOPATH by concatenate $GOPATH and "go-guess-gopath"
  result, use it with gofmt-command to format code.

Then working on 2 go buffers from different projects (with different vendor directories). The `gofmt` can't get the correct GOPATH correctly unless you call `go-set-project` after jump between buffers. 

The dynamic way can create correct GOPATH for current call to `gofmt-command` without interfering other projects.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1135,7 +1135,8 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
         (errbuf (if gofmt-show-errors (get-buffer-create "*Gofmt Errors*")))
         (coding-system-for-read 'utf-8)
         (coding-system-for-write 'utf-8)
-        our-gofmt-args)
+        our-gofmt-args
+	gofmt-full-cmd)
 
     (unwind-protect
         (save-restriction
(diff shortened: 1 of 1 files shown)


go-mode.el
```diff
@@ -1156,11 +1157,14 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
           (setq our-gofmt-args (append our-gofmt-args
                                        gofmt-args
                                        (list "-w" tmpfile)))
-          (message "Calling gofmt: %s %s" gofmt-command our-gofmt-args)
+
+	  (setq gofmt-full-cmd (format "GOPATH=%s:%s %s %s" (go-guess-gopath) (getenv "GOPATH") gofmt-command (mapconcat 'identity our-gofmt-args " ")))
```
syohex: This command format(+ `shell-command`) works only on `sh`, `bash`, `zsh`(`shell-command` uses `SHELL` environment variable shell for executing command). This raises error on `csh`, `tcsh`, `fish`(and Windows). I suppose that you should use `process-environment` variable for setting environment variable temporarily.

See
- https://www.gnu.org/software/emacs/manual/html_node/elisp/System-Environment.html

dominikh:
Thanks, but I don't think I'll merge this. One design goal when writing go-set-project was that go-mode wouldn't try to automatically update your GOPATH when changing buffers. People should be able to overwrite the whole GOPATH (e.g. via setenv), and your change makes that impossible. Furthermore, we don't want to modify all of our functions that invoke external commands, and there already exist third party packages that do not do this, either. It would be very awkward if some commands guessed the GOPATH, while others didn't. Also, merging the guessed GOPATH with the already set one isn't always the right solution, either. For example when having per-customer GOPATHs with strict vendoring requirements, you do not want your global GOPATH to be mixed with the customer's GOPATH.

The recommended solution is to automatically call go-set-project in a way that suits your workflow. Some people use directory variables, some people use projectile's hooks, some people just call it manually. There is no 100% perfect solution, which is why we let people build their own solutions. Ultimately, you could also `advise` the commands that you care about, in your own Emacs configuration.

PR #139, opened 2016-06-01
Enable Go syntax highlighting in godoc-mode

kostya-sh:


kostya-sh:
I am a bit unsure about this change myself but decided to send
the PR anyway to get some feedback.

I do beleive that it makes go docs easier to
read (especially "godoc package" output) but it certainly
highlights words that shouldn't be highlighted.

Would you consider including godoc syntax highlighting (maybe
behind a customizable variable)?

Maybe the hightling can be made more accurate by providing
godoc-mode specific hightling rules. Is it even possible given
that all 3 godoc commands are not very consistent in a way they
format the output?

dominikh:
This has been rejected before in #87. To reiterate the main point from that PR: Syntax highlighting is harmful if it's incorrect. And this hack of adding highlighting to godoc is incorrect too often.

If this is something that individual users want to endure, they should just add a hook to their Emacs configuration. It's not something I want to encourage in go-mode, not even behind an option.

PR #140, opened 2016-06-30
Fix non-existing function being used in go-unused-imports-lines

kostya-sh:
I've noticed that non-existing buffer-file-truename function is used
while browsing the source code. I beleive buffer-file-name should be
used instead. I haven't tested the change though.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1474,7 +1474,7 @@ archive files in /pkg/"
                               (string-to-number error-line-num)))))
                     (split-string (shell-command-to-string
                                    (concat go-command
-                                           (if (string-match "_test\\.go$" buffer-file-truename)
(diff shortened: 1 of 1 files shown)


dominikh:
```
buffer-file-truename is a variable defined in `C source code'.
Its value is "~/notes/foo.txt"
Local in buffer foo.txt; global value is nil

  Automatically becomes permanently buffer-local when set.
  This variable may be risky if used as a file-local variable.

Documentation:
Abbreviated truename of file visited in current buffer, or nil if none.
The truename of a file is calculated by `file-truename'
and then abbreviated with `abbreviate-file-name'.
```

kostya-sh:
Whoops, `buffer-file-truename` is a variable, no wonder I could not find it! Sorry for the bogus PR.

PR #143, opened 2016-07-10
[master] Create and remove buffer in-place.

hotpxl:
Instead of reusing the old buffer, `godef` now creates a new buffer and deletes it after use. So I won't get a useless `*godef*` when I list all buffers.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1527,25 +1527,25 @@ description at POINT."
       (error "godef does not reliably work in XEmacs, expect bad results"))
   (if (not (buffer-file-name (go--coverage-origin-buffer)))
       (error "Cannot use godef on a buffer without a file name")
-    (let ((outbuf (get-buffer-create "*godef*"))
+    (let ((outbuf (generate-new-buffer "*godef*"))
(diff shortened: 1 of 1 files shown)


dominikh:
Thanks, but I don't think this change is worthwhile. Gocode, guru and some other packages will behave the same way. I suggest filtering your list of buffers instead to hide, or demote, special buffers (those surrounded by asterisks).

hotpxl:
`company-go` which uses Gocode creates and deletes buffers on the fly. [Here](https://github.com/nsf/gocode/blob/master/emacs-company/company-go.el#L57) for a reference.

hotpxl:
`company-go` code I linked to is part of the official Gocode release.
Guru also cleans up temporary buffers ([here](https://github.com/golang/tools/blob/master/cmd/guru/go-guru.el#L176)). Although I don't use it.

dominikh:
I'll reconsider and possibly merge the PR later today.

hotpxl:
Hi. Just want to check on the status. Has anything evolved since?

PR #149, opened 2016-09-08
Enabled cover.out as coverage file default

nanjj:
`go-test-current-coverage` in `go-test` is using `cover.out` as default
coverage file, after we enabled in `go-coverage`, we can just type enter
to reuse it.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -146,6 +146,14 @@ point to the wrapper script."
   :type 'string
   :group 'go)
 
+(defcustom go-packages-command go-command
+  "The command to list go packages.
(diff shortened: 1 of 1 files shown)


dominikh:
What are these `go-test-current-coverage` and `go-test` you are referring to? Neither are functions in go-mode.

nanjj:
`go-test-current-coverage`  is from https://github.com/nlamirault/gotest.el . If `go-coverage` has a default coverage file like `cover.out`, it's easier to use `go-coverage` with `go-test-current-coverage`.

PR #150, opened 2016-09-12
Fix go-goto-arguments when in subword-mode

dspereira004:
When calling `go-goto-arguments` in a CamelCase function with at least two words, the point doesn't go to the first argument.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1835,7 +1835,7 @@ If ARG is non-nil, anonymous functions are skipped."
 If ARG is non-nil, anonymous functions are skipped."
   (interactive "P")
   (go-goto-function-name arg)
-  (forward-word 1)
+  (forward-word-strictly 1)
   (forward-char 1))
 
 (defun go--goto-return-values (&optional arg)


go-mode.el
```diff
@@ -1835,7 +1835,7 @@ If ARG is non-nil, anonymous functions are skipped."
 If ARG is non-nil, anonymous functions are skipped."
   (interactive "P")
   (go-goto-function-name arg)
-  (forward-word 1)
+  (forward-word-strictly 1)
```
dominikh: This function is not defined in my GNU Emacs 24.5.1. Where is it from?

syohex: https://lists.gnu.org/archive/html/emacs-diffs/2016-02/msg00108.html

Its function was introduced by above commit at 2016/FEB. Only Emacs 25 RC or development Emacs users can use it now.

dominikh: Well, the Emacs folks clearly have gone insane. I will fix this myself, in a backwards compatible way. *Someone* has to be backwards compatible, right? :-)

dspereira004: Haha that's right. Thanks for fixing this!

dominikh:
Closing in favour of #167

PR #157, opened 2016-10-23
Make go-guru-build-tags a list of strings

### LINKED ISSUE #156: go-guru: go-guru-build-tags in go-guru.el does not work
The `go-guru-build-tags` customize variable doesn't work properly.  To reproduce the crash I'm seeing, perform the following steps:
- Open Emacs with `go-guru.el` loaded
- Type `M-x customize-group RET go-guru RET`
- Type `M-x eval-expression RET (setq go-guru-build-tags "one") RET`
- Type `M-x toggle-debug-on-error RET`
- Run any go-guru command that calls `go-guru--command`

The following stacktrace is produced:

```
Debugger entered--Lisp error: (wrong-type-argument sequencep 111)
  mapconcat(identity "one" ",")
  (format "-tags=%s" (mapconcat (quote identity) go-guru-build-tags ","))
  eval((format "-tags=%s" (mapconcat (quote identity) go-guru-build-tags ",")) nil)
  eval-expression((format "-tags=%s" (mapconcat (quote identity) go-guru-build-tags ",")) nil)
  call-interactively(eval-expression nil nil)
  command-execute(eval-expression)
```

Note that the stacktrace is only produced if `go-guru-build-tags` is a non-empty string, which explains why this bug was not immediately noticed.

I see a few possible solutions.
1. Change the following line in `go-guru--command` from

```
(format "-tags=%s" (mapconcat 'identity go-guru-build-tags ","))
```

back to just

```
"-tags" go-guru-build-tags
```

If choosing this solution it would probably be good to expand the docstring for `go-guru-build-tags` to indicate that the value must be comma-separated when including multiple build tags.
1. Change the declaration for `go-guru-build-tags` to a list of strings

```
(defcustom go-guru-build-tags '()
  "Build tags passed to guru."
  :type '(repeat string)
  :group 'go-guru)
```

and leave `go-guru--command` as it currently is.  This solution seems better except that it would break anyone currently setting `go-guru-build-tags` as a string.  Therefore, I'm not sure which one should be preferred.

nwidger:
Previously, go-guru-build-tags was a single string.  This commit fixes a stack trace generated by `go-guru--command` if
go-guru-build-tags is set to any value other than the default empty
string.

Fixes #156

### DIFF
--- a/go-guru.el
+++ b/go-guru.el
@@ -70,9 +70,9 @@
   nil
   "History of values supplied to `go-guru-set-scope'.")
 
-(defcustom go-guru-build-tags ""
+(defcustom go-guru-build-tags '()
   "Build tags passed to guru."
-  :type 'string
+  :type '(repeat string)
   :group 'go-guru)
 
 (defface go-guru-hl-identifier-face


dominikh:
Thanks!

PR #158, opened 2016-10-31
use (goto-char (point-min)) instead of (beginning-of-buffer)

aborn:
1. go-guru.el:243:12:Warning: `beginning-of-buffer' is for interactive use only;
   use`(goto-char (point-min))' instead.
2. code format


dominikh:
Hi, I've cherry-picked the code change. The code formatting will be fixed in a separate commit in the future. 

As a general bit of advice for PRs: try avoiding having code changes and formatting changes in the same commit. It makes them very hard to review.

PR #161, opened 2016-11-18
Use file-truename to make it works with symlinks

chmouel:
Fixes https://github.com/golang/go/issues/16219


alandonovan:
I would rather we fix the bug in the right place than work around it in Emacs---and since this is a work-around, it should be properly identified as such, with a link to the underlying bug.

dominikh:
If I understand correctly, the bug is now fixed in the tools, so I am going to close this PR.

PR #167, opened 2016-12-06
use forward-word-strictly when available

nverno:
This is just a follow up to the previous pull request.  It just
introduces a simple macro to use forward-word-strictly when 
it is bound.

`go-goto-function-name` and the like don't work properly in 
emacs >= 25 with `subword-mode` when types are camelCased,
eg,

```
func (list *LinkedList) Sum() int {
	// point
}
```

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -50,6 +50,12 @@
 (defmacro go--has-syntax-propertize-p ()
   (boundp 'syntax-propertize-function))
 
+(eval-when-compile
+  (defmacro go--forward-word (&optional arg)
+   (if (fboundp 'forward-word-strictly)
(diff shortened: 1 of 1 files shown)


dominikh:
Thanks for the PR. Two things:

1. Please call the macro `go--forward-word`, i.e. use two dashes to signal that this is an internal symbol 
2. What's the significance of eval-when-compile here? If it's just an optimization, I think we can do without here. The body is small, and none of the other macros are wrapped in eval-when-compile.

nverno:
if you use `eval-when-compile` then that macro will be expanded during compilation so the `go-forward-symbol` wont be exported when using the library, since it isn't needed anywhere else.

PR #168, opened 2016-12-22
Don't popup a new emacs frame for godoc

mackong:


petergardfjall:
Any chance of seeing this PR merged..? :)

psanford:
I'm a bit unsure of the specific issue here. Functionally testing this I don't see any behavioral differences before vs after using my normal godoc workflows. The docs for `(display-buffer)` say the following for this argument:

```The ACTION argument can also have a non-nil and non-list value.
This means to display the buffer in a window other than the
selected one, even if it is already displayed in the selected
window.```

That seems to say the only thing this affects is if you already have a buffer called "*godoc time*" open and selected and you run `(godoc "time")` again it will reopen the help buffer in a separate frame again. Is that what you are trying to fix here?

Maybe attaching short clips of what before and after this diff is applied looks for you would help clarify what we are trying to fix here.

petergardfjall:
I'm not 100% sure but I believe this PR fixed behavior seen when running together with the neotree project explorer. https://github.com/dominikh/go-mode.el/issues/180#issuecomment-369322570

psanford:
I can't reproduce the issue described with neotree. I do see different behavior depending on the size of my emacs window but not the issue described there.

psanford:
There has been no action on this PR for 6 months therefore I'm going to close it.

If someone can provide a reproducible example of an issue that this fixes we can reopen it.

PR #173, opened 2017-01-16
Make `C-u C-c C-j' show a Go definition in another window

neiljerram:
I think this is a simple and very useful usability enhancement - WDYT?

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1496,8 +1496,9 @@ description at POINT."
     (file-error (message "Could not run godef binary"))))
 
 (defun godef-jump (point &optional other-window)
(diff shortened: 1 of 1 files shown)


dominikh:
We already have `C-x 4 C-c C-j` mapped to `godef-jump-other-window`, `C-x 4` being the idiomatic binding for "other window". I don't think we should use C-u for this. We may want to use it for some other toggle in the future, and there's already a canonical binding for this.

neiljerram:
Fair enough.  I _thought_ I looked for a binding for `godef-jump-other-window` (with `C-h w`) and _didn't_ find one, before trying this code change - but I do see that `C-x 4` binding now.

PR #178, opened 2017-01-26
Fix compiler warnings under Emacs 25.1

phst:


go-mode.el
```diff
@@ -493,13 +494,13 @@ STOP-AT-STRING is not true, over strings."
...
     (if (and (save-excursion (beginning-of-line) (go-in-string-p))
-             (looking-back "`")
+             (= (char-before) ?`)
              (not stop-at-string))
         (backward-char))
     (if (and (go-in-string-p)
              (not stop-at-string))
         (go-goto-beginning-of-string-or-comment))
-    (if (looking-back "\\*/")
+    (if (looking-back "\\*/" (line-beginning-position))
```
dominikh: Is there actually a performance gain here, considering we're looking for a static string?

go-mode.el
```diff
@@ -527,7 +528,7 @@ STOP-AT-STRING is not true, over strings."
         (save-excursion
           (beginning-of-line)
           (go--backward-irrelevant t)
-          (setq val (looking-back go-dangling-operators-regexp))
+          (setq val (looking-back go-dangling-operators-regexp (- (point) 2)))
```
dominikh: We should probably introduce a constant for the magic value of 2.

go-mode.el
```diff
@@ -582,7 +583,8 @@ current line will be returned."
         (if (go-previous-line-has-dangling-op-p)
             (- (current-indentation) tab-width)
           (go--indentation-for-opening-parenthesis)))
-       ((progn (go--backward-irrelevant t) (looking-back go-dangling-operators-regexp))
+       ((progn (go--backward-irrelevant t)
+               (looking-back go-dangling-operators-regexp (- (point) 2)))
```
dominikh: Same as above.

go-mode.el
```diff
@@ -1503,7 +1506,9 @@ description at POINT."
         (if (not (godef--successful-p file))
             (message "%s" (godef--error file))
           (push-mark)
-          (ring-insert find-tag-marker-ring (point-marker))
+          (if (eval-when-compile (fboundp 'xref-push-marker-stack))
+              (xref-push-marker-stack)
```
dominikh: Not in the scope of this PR, but we should probably hook into xref properly, so that `xref-find-definitions` works as expected.

go-mode.el
phst: I haven't benchmarked it, but Emacs now soft-requires the limit (i.e. you get a warning if you don't specify it), and the docstring recommends to specify a non-nil limit.

go-mode.el
phst: Done.

go-mode.el
phst: Done.

phst:
I've address the comments.

dominikh:
Merged as d511e352c81ccdc08271024104d567b3bd15bf67 and 5056b857a90d732d78162b3263a1dd94f7c5afe3 – thanks for the PR and sorry for the delay in merging it.

PR #182, opened 2017-03-13
Fixed filename processing on Windows

AdVocem:
Fixed behaviour on Windows systems while splitting the filenames by ":" that results into an errors.

### DIFF
--- a/go-guru.el
+++ b/go-guru.el
@@ -229,7 +229,8 @@ output of the Go guru tool."
 	    (setq start end) ; break out of loop
 	  (setq p (1- p)) ; exclude final space
 	  (let* ((posn (buffer-substring-no-properties start p))
-		 (flen (cl-search ":" posn)) ; length of filename
+             (file-line-pos (go-guru--split-to-file-line-pos posn))
+             (flen (length (car file-line-pos))) ; length of filename
 		 (filename (if (< flen 19)
 			       (substring posn 0 flen)
 			     (concat "…" (substring posn (- flen 19) flen)))))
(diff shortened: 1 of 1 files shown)


AdVocem:
- All changes are related, except indentation on line 224 (will not be fixed)
- godef--find-file-line-column does another thing in a much heavier way, so I'll leave that unchanged too

dominikh:
> Customizable variable was introduced to avoid magical numbers in code

Which is fine, but entirely unrelated to "fixing file name processing on Windows" and has no place in this PR or commit. (It should also have been a constant, not a customizable variable; we don't want to expose it to the user at this point in time)

> All changes are related, except indentation on line 224, but originally it was a crap because of mixed tab and space symbols in a one function (will not be fixed back)

Which is fine, but also unrelated. It should at least be its own commit that does nothing but change the formatting.

> godef--find-file-line-column does another thing in a much heavier way. Currently reassembling of the file:line:col is executed only if problem exists, so I'll leave that unchanged too

Your solution is a hack, which works around a bug in the original code (blindly splitting on ":" is inherently wrong. Non-windows file systems permit colons anywhere in the file name, so both the original code and your fix are insufficient.)

godef--find-file-line-column is much closer to a sane implementation. It is indeed not a perfect drop-in replacement, though, as it operates on columns, whereas guru emits byte positions. We should probably split the parsing of `file:line:col` into a separate function, then reuse that one both in godef--find-file-line-column and go-guru, interpreting the column as required. Or, ideally, Emacs already has a public function that does this for us.

> If you have a different opinion, please, clarify what you exactly want

I've been rather clear about what I want, you just chose to ignore me. Will you be able to follow my requests, or should I just implement the change myself?

AdVocem:
Thanks for clarifiing but it seems that I have no so much free time to support legacy formatting

> we don't want to expose it to the user at this point in time

Strange point of view. Do we both talking about  lisp code?

> Your solution is a hack

I know that but that is better than nothing. 

> or should I just implement the change myself?

I am not a telepath so if you know a better solution you are free to implement it. I have no objections.
I just want to have a solution publicly available.

AdVocem:
I finally rewrite my code to satisfy your requirements. This hack fixes the problem for the 99.9% of users so it could be considered as temprorary solution untill further refactoring.

dominikh:
Thanks, I'll take a look soon!

PR #185, opened 2017-04-06
Fix off-by-one in point calculations

Wilfred:
Emacs point is 1-indexed rather than 0-indexed. As a result, calling
`godef-jump` on the following code:

```go
 foo|(bar) // assuming | is point
```

would find the definition of `bar` rather than `foo`.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1478,7 +1478,8 @@ description at POINT."
                                "-f"
                                (file-truename (buffer-file-name (go--coverage-origin-buffer)))
                                "-o"
-                               (number-to-string (position-bytes point)))
+                               ;; Emacs point and byte positions are 1-indexed.
+                               (number-to-string (1- (position-bytes point))))
           (with-current-buffer outbuf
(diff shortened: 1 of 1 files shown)


Wilfred:
@dominikh ping, are you happy with this?

dominikh:
Hi.

I don't think the patch is correct. You're decrementing point, not the computed byte offset. This is wrong by simple deduction: If point is at the beginning of the buffer, point - 1 will be 0, and (position-bytes 0) is nil.

I wasn't aware that godef's offset is 0-based, but you seem to be right. We should decrement the result of position-bytes.

Wilfred:
@dominikh oh, nice catch. I've updated the PR.

I didn't seen any references to 1-based indexing in the docstring or source code for `position-bytes`, but the Emacs manual says:

> Buffer positions are measured in character units. This function returns the byte-position corresponding to buffer position position in the current buffer. This is 1 at the start of the buffer, and counts upward in bytes.
>
> -- https://www.gnu.org/software/emacs/manual/html_node/elisp/Text-Representations.html

dominikh:
Merged, thank you.

PR #188, opened 2017-05-01
Fix all compilation warnings in go-guru.el

phst:


go-guru.el
```diff
@@ -117,7 +117,7 @@
   (define-key m ">" #'go-guru-callees)
   (define-key m "x" #'go-guru-expand-region)) ;; x for expand
 
-(define-key go-mode-map (kbd "C-c C-o") #'go-guru-map)
+(define-key go-mode-map (kbd "C-c C-o") 'go-guru-map)
```
dominikh: What's the warning that triggered this change?

go-guru.el
```diff
@@ -450,6 +453,22 @@ highlights from previously highlighted identifier."
   (go-guru-unhighlight-identifiers)
   (go-guru--hl-identifier))
 
+;;;###autoload
```
dominikh: Why was this chunk of code moved around?

go-guru.el
phst: "Function might not be defined". Indeed `go-guru-map` isn't a real function that could be called, but a keymap.

dominikh: Right, of course.

PR #189, opened 2017-05-11
gopkgs support

firstrow:
Since my emacs go-import-add started to response slowly I found this way to improve performance. 
gopkgs works way much faster compared to native and go list funcs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,9 @@ the following extra features to provide an improved experience:
   - A function for jumping to the file's imports (`go-goto-imports` -
     `C-c C-f i`)
   - A function for adding imports, including tab completion
-    (`go-import-add`, bound to `C-c C-a`)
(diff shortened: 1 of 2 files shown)


dominikh:
The implementation of gopkgs is identical to that of go-packages-native: it looks through GOPATH/pkg for installed packages. Instinctively, I find it difficult to believe that the two should have any significant difference in performance; and if they do, we should probably improve go-packages-native, instead of relying on an external tool.

Have you tested both implementations with a hot file system cache, or was gopkgs simply faster because you tested it after you had already run go-packages-native? Also, can you provide concrete numbers? Thanks!

firstrow:
@dominikh sure, I'll do the math and post result here.

firstrow:
sorry, I have no free time for this now.

PR #196, opened 2017-06-08
Evaluate error conditions before prompting for new name in go-rename

helgeblod:
It's super annoying to get the error "Please save the current buffer before invoking go-rename" after creating a new name just to save and do everything all over again. This PR fixes this by returning the error immediately before prompting for a new name if the buffer is not saved or the buffer is not a file buffer. 

To test, open a file, edit something *without* saving and then try to invoke `go-rename`. The error should then pop-up immediately without the prompt for a new name. Saving the file and retrying `go-rename` will make it behave as expected.

### DIFF
--- a/go-rename.el
+++ b/go-rename.el
@@ -40,8 +40,10 @@
   "Rename the entity denoted by the identifier at point, using
 the `gorename' tool. With FORCE, call `gorename' with the
 `-force' flag."
-  (interactive (list (read-string "New name: " (thing-at-point 'symbol))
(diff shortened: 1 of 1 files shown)


dominikh:
Thanks for the contribution. The idea sounds fine to me, I'll review the implementation as soon as I find the time.

helgeblod:
Cool 😀 My elisp skills are limited, so don't be afraid to give me some honest feedback 👍

PR #199, opened 2017-06-12
Add go-present-mode: edit Go present files.

LMMilewski:
Go-present-mode is a mode for viewing and editing Go present files
(mainly slides). It strives to provide view and features consistent with
Go present tool web view.


go-present-mode.el
```diff
@@ -0,0 +1,1278 @@
...
+  "Regular expression matching .html command.
+
+Subexpression 1 is the file name.")
+
+(defconst go-present--code-start-regexp "^\\.\\(code\\|play\\)\\s-+"
+  "A ragular expression matching start of a code section.
+
+Subexpression 1 is the command (.code or .play).")
+
+(cl-defun go-present-open-at-point ()
```
dominikh: Is there a way this routine could be written in terms of thing-at-point, either leveraging it, or adding functionality to thing-at-point?

LMMilewski:
>  I'll not review it all in detail; instead I'll use the mode and look for user-facing bugs. 

Sounds good.

> The overall code quality looks very clean, however.

Thank you

> I did see that you're using fontification information to parse syntax. That's pretty much a no-go, 

I admit that using font-lock here is not ideal (e.g. I had to write elisp to handle copy-paste in code sections, spend time on supporting undo operations, etc.). What is a better way to do this?

In most cases (all?) I need to both parse the text and update faces. For example, I store a property with url on links like [[www.example.com][example]] because I already need a function for fontification (only "example" part stays visible) and reading a property is less likely to be buggy than implementing another link parser for open-at-point.

I'm actually one of the people who disable fontification (at least in programming modes), but I'm not sure that it makes sense for go-present-mode. What this mode provides is, for the most part, showing the text in a similar way to the present tool so that one doesn't have to switch to the browser all the time. Without font-lock it's not that useful.

I'm happy to update the code if you can clarify what the alternative is.

go-present-mode.el
LMMilewski: I think that it can, but that won't simplify the code.
Currently I can do a regexp match and then pick a submatch as the result (and give it to a function like browse-url).
The alternative is to call thing-at-point (after I define forward/backward functions), do a regexp match on the result, and then pick submatch.

I can update the code if you think that thing-at-point is a better approach (it might have benefits that I'm not aware of).

go-present-mode.el
```diff
@@ -0,0 +1,1278 @@
...
+  "Base port to serve Go present on.
+
+Go present serves on the first free port that is not smaller than
+`go-present-base-port'."
+  :type 'integer
+  :group 'go-present)
+
+(defface go-present-bold-face '((t (:inherit bold)))
+  "Face for bold text outside code sections.")
+(defvar go-present-bold-face 'go-present-bold-face)
```
LMMilewski: It doesn't work without defvars. When I wrote the code, I spent some time researching this and couldn't figure out how to implement fontification without defvars in this case (if you comment out defvars then nothing is fontified). In the end I decided that it wasn't worth the time to look into this further.

If you know how to fix it, then I'm happy to update the code. In fact, I would really appreciate the feedback because I wasted so much time on this (and I hope that it wasn't me just missing something obvious).

I suspect that I'd have to update references to those faces in go-present--keywords definition, but I'm not sure how.

LMMilewski: @phst  showed me how to fix this. I uploaded a new version which includes the change.
It boiled down to quoting those names properly. I must've had something else broken when I tried this before.

dominikh:
> I admit that using font-lock here is not ideal (e.g. I had to write elisp to handle copy-paste in code sections, spend time on supporting undo operations, etc.). What is a better way to do this?

In general, you parse the surrounding context on an as-needed basis. Particularly expensive operations can be cached in buffer local variables. Look at how go-mode does indentation for an example. We look back as much as we need to determine the correct indentation, and cache the information. Most of the time, you only need to scan a small region so that doing ad hoc parsing is cheap.

Something like link parsing is an interesting case. If you linkify the buffer contents, you can certainly store the URL in a property. But you'd also want to provide a function a la `go-present-open-url` that doesn't depend on fontification and instead parses the URL under point.

> I'm actually one of the people who disable fontification (at least in programming modes), but I'm not sure that it makes sense for go-present-mode. 

I assume that there are still useful features that don't depend on fontification? Paren matching, indentation, functions for manipulating slides? That aside, even if fontification is enabled, it's not a good idea to store parse trees in properties. As you've found out yourself, it requires a lot of special care (copy&paste, ...) and is prone to breakage.

Finally, I think it's best if you maintain this mode yourself. I would've quite liked to merge it into go-mode, but I'm lacking the time to maintain another large chunk of code and I don't want to slow you down (any longer...) by taking forever to review PRs.

I'd be more than happy to add a reference to your mode in the README and documentation, though.

LMMilewski:
Sounds good. I'll move the code to its own repo. Thank you for feedback and suggestions.

I think that we're on the same page when it comes to indentation and similar use-cases. I'll look into support for no-font-lock use-case, but probably only where it won't require maintaining multiple versions of parsing logic.

PR #201, opened 2017-06-25
adding go-alternate to go-mode.el

c25l:
Hey,
This is a useful convenience function for switching between files and their tests. If there is anything wrong stylistically here, I'd be happy to make changes.

Thanks!
Chris

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -446,6 +446,7 @@ For mode=set, all covered lines will have this weight."
     (define-key m (kbd "C-x 4 C-c C-j") #'godef-jump-other-window)
     (define-key m (kbd "C-c C-d") #'godef-describe)
     (define-key m (kbd "C-c C-f") 'go-goto-map)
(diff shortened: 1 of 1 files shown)


dominikh:
Thank you, but go-mode already supports this feature via `ff-find-other-file`. See the `go-other-file-alist` variable for details.

c25l:
Yeah, i see the implementation now, and it really works. I cannot, however, find any documentation about it anywhere else in go-mode.el, nor is there a keybinding for it. Especially for the "I'm new to {E|Space}macs" crowd, it might be nice to have either/both. I would be willing to add this, if you think it's valid.

c25l:
Also thanks for taking the time to look over my PR. I really like using go-mode.el.

dominikh:
You can send a PR for the readme if you'd like, possibly in the list of standard Emacs features that we  support, we already list things such as beginning-of-defun.

On a broader scope, however. people should learn Emacs, not go-mode.

PR #204, opened 2017-07-20
add go-occur-definitions function

tt67wq:
Add a function to quick view all the functions and type in a go file
I think that's quit useful for me, and I add a key bind for this 'C-c C-o'

you can have a try -.-

forgive my poor English


psanford:
Thanks for the pr. I don't think this function will be generally useful to enough people to justify adding it to go-mode.

PR #207, opened 2017-08-14
Fix removal of unused imports on remote system using TRAMP

thatnerdjosh:


dominikh:
If we do this for removing unused imports, don't we also have to do it for functions that use godef? And possibly others?

NerdsvilleCEO:
There are other functions with don't work with TRAMP I believe, would need to look, but I don't believe this breaks anything already working, correct me if I'm wrong

dominikh:
People should use LSP/gopls for this functionality, which I'm told already works with TRAMP.

PR #208, opened 2017-08-25
Skip spaces at the point before jumping to a definition

arkdro:
If there are spaces in the current line between the point and a function, then skip these spaces and then jump to the definition of the function.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1516,7 +1516,9 @@ description at POINT."
   "Jump to the definition of the expression at POINT."
   (interactive "d")
   (condition-case nil
-      (let ((file (car (godef--call point))))
+      (let* ((delta (skip-chars-forward "[:space:]"))
+             (new-point (+ point delta))
+             (file (car (godef--call new-point))))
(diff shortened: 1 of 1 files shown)


dominikh:
If we do this for godef-jump, we also need to do it for all other functions that look at symbols. That's at least godef-describe, possibly the functions in go-guru, possibly godoc. 

What's the use case for this change?

arkdro:
The use case is: after you scrolled the buffer, the cursor is somewhere at the beginning of a line.
Without the patch you need to move the cursor to the function name and then you are able to jump to a definition.
With this patch you can jump to the definition without extra key presses.

For this patch I was thinking about godef-jump only. I can take a look at godef-describe as well. About the others - I'm just not sure, if it is necessary.

dominikh:
I'm tempted to do nothing and recommend using `back-to-indentation` (bound to `M-m` by default, IIRC), which is only one extra key press and works universally, not just for godef-jump.

arkdro:
all right then, closing the PR

PR #210, opened 2017-09-01
Add a new customization variable to allow adding additional flags to …

phst:
…gorename

### DIFF
--- a/go-rename.el
+++ b/go-rename.el
@@ -35,6 +35,12 @@
   :type 'string
   :group 'go-rename)
 
+(defcustom go-rename-flags ()
+  "Additional flags for the ‘gorename’ binary."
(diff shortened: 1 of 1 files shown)


dominikh:
Which additional flags would you pass to gorename with this?

phst:
There are a couple of generic flags such as `-v` that are currently not covered by other options.

dominikh:
The flags that are not supported are `-from`, `-d`, and `-v`. Using either `-from` or `-d` would break the function, leaving just `-v`. It would make more sense to add an option specifically for verbose logging.

dominikh:
Got no response, and gorename isn't all that useful anymore with gopls. Closing.

PR #211, opened 2017-09-18
move to godef result line/column by with-current-buffer

aki2o:
When finished to `find-file` for godef result file, It's not sure that `current-buffer` is the file.
Therefore, we should use `with-current-buffer` when move to the line and column of godef result.


dominikh:
Please see 498452d560c07e308e493b7ea7467880dfe56987 and #85 for why this change is incorrect.

PR #212, opened 2017-09-25
Treat _ as a symbol constituent

Wilfred:
This allows users to navigate in FOO_BAR by word. Previously, it would
move across the whole FOO_BAR.

This is recommended by Stefan Monnier, the former GNU Emacs
maintainer:
https://emacs.stackexchange.com/questions/1075/whats-the-difference-between-words-and-symbols#comment2567_1077

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -386,8 +386,7 @@ For mode=set, all covered lines will have this weight."
     (modify-syntax-entry ?\' "\"" st)
     (modify-syntax-entry ?`  "\"" st)
     (modify-syntax-entry ?\\ "\\" st)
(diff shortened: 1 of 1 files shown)


Wilfred:
@dominikh friendly ping -- this keeps words and symbols distinct. Would you be willing to merge this?

dominikh:
This would also require changing some of the regular expressions, which currently only match on word characters, not symbols. And that would lead to the problems I've mentioned in https://github.com/dominikh/go-mode.el/pull/69

psanford:
I'm going to close this PR as I think its a duplicate of #69.

PR #213, opened 2017-09-28
remove go-mode-autoloads because it will be generated during installa…

gnufied:
Fixes https://github.com/melpa/melpa/issues/5019


dominikh:
The file exists because of historical reasons (the original go-mode that we replaced included the file, unfortunately.)

But at this point, the presence of the file is probably causing more harm than good.

gnufied:
This PR will not be enough to fix the bug though. I think we will need `;;;###autoload` comments in `go-mode.el` too.

dominikh:
go-mode.el has plenty of autoload cookies, and to the best of my knowledge they cover all the functions that are worth autoloading. Did we miss any?

FTR, the existing go-mode-autoloads file is generated from the autoload cookies, it isn't hand-written.

tarsius:
> FTR, the existing go-mode-autoloads file is generated from the autoload cookies, it isn't hand-written.

So you should be good to go. (Consider this a friendly ping.)

dominikh:
@tarsius there we go.

tarsius:
Thanks!

PR #219, opened 2017-11-15
Include the column number in the parsing for unused imports

noonie2k:
I've found that the output of `go build` includes a column number as part of the output and so `go-remove-unused-imports` is failing to match the output for the given regexp.

I've included and additional `:[[:digit:]]` check in the regexp which is now matching my `go build` output and removing imports.

FYI; I'm using go version 1.9.2

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1408,7 +1408,7 @@ It looks for archive files in /pkg/."
   (reverse (remove nil
                    (mapcar
                     (lambda (line)
-                      (when (string-match "^\\(.+\\):\\([[:digit:]]+\\): imported and not used: \".+\".*$" line)
+                      (when (string-match "^\\(.+\\):\\([[:digit:]]+\\):[[:digit:]]: imported and not used: \".+\".*$" line)
                         (let ((error-file-name (match-string 1 line))
                               (error-line-num (match-string 2 line)))
                           (if (string= (file-truename error-file-name) (file-truename buffer-file-name))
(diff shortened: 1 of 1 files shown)


dominikh:
1. Column numbers have only recently been added to the output, so we need to make them optional
2. Your modified regexp only permits a single digit, it should allow many digits

noonie2k:
Good calls; in my haste I'd missed these points.

I've updated the regex and tested with the following test input.

```"# command-line-arguments
./go.go:5:2: imported and not used: \"bufio\"
./go.go:10:2: imported and not used: \"net/http\"
./go.go:20:2: undefined: fmt.Prinln
"

"# command-line-arguments
./go.go:5: imported and not used: \"bufio\"
./go.go:10: imported and not used: \"net/http\"
./go.go:20: undefined: fmt.Prinln
"```

The updated regex found both of the example outputs, highlighting just the file names and line numbers for the capture groups in both cases, and ignored the `undefined` error.

Executing the command in my test program successfully removed the two offending lines.

**Comment by noonie2k** [comment]
Any update on this?

File: go-mode.el [hunk 1]
```diff
@@ -1408,7 +1408,7 @@ It looks for archive files in /pkg/."
   (reverse (remove nil
                    (mapcar
                     (lambda (line)
-                      (when (string-match "^\\(.+\\):\\([[:digit:]]+\\): imported and not used: \".+\".*$" line)
+                      (when (string-match "^\\(.+?\\):\\([[:digit:]]+\\):\\([[:digit:]]+:\\)? imported and not used: \".+\".*$" line)
```
dominikh: Is there a specific reason you switched from `.+` to `.+?` for the file name? That shouldn't be necessary here as far as I am aware.

psanford:
This was fixed in ac6dd8c1f4b969fb93ecb8ac62b0700eeb1b62fd

PR #223, opened 2017-11-29
Enhance imenu support

brantou:
Use [go-outline](https://github.com/lukehoban/go-outline) to enhance imenu support, and automatically downgrade to the original implementation when the go-outline is unavailable or any error occurs.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -987,10 +987,11 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
   ;; ff-find-other-file
   (setq ff-other-file-alist 'go-other-file-alist)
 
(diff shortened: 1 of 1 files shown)


brantou:
@dominikh Do you think it is not necessary? Or my approach is not good. I hope you can give me feedback.

dominikh:
Sorry, but I didn't have the chance to look at this yet. I'll probably be able to check it out some time in February, at which point I'll certainly provide feedback.

dominikh:
We'll have to postpone this. At this point in time, I am not looking to add more external dependencies to go-mode. We'll cover this in the future when go-mode is redesigned from the scratch up, likely relying on LSP by then.

PR #225, opened 2018-01-07
recommend a hook that doesn't break autoloading

rski:
Instead of recommending a hook that runs for every mode and exits if
the current mode isn't go-mode, recommend a hook that runs for go-mode
and adds a local before-save-hook that runs gofmt


rski:
This actually breaks things if the file's mode is changed because the hook will persist.

Issue #218, opened 2017-10-26
go-guru doesn't work on Windows


seagle0128:
Windows 10, GNU Emacs 25.3, Go 1.9.1

`go-guru` returns the error below, but `godef` works well.

```elisp
Debugger entered--Lisp error: (wrong-type-argument integer-or-marker-p nil)
  go-guru--goto-byte-column(256)
  go-guru--goto-pos("C:\\Go\\src\\fmt\\print.go:256:6" nil)
  go-guru-definition()
  funcall-interactively(go-guru-definition)
  call-interactively(go-guru-definition nil nil)
  command-execute(go-guru-definition)
```

seagle0128:
Seems `go-guru--goto-pos` doesn't handle the file path like "C:\\Go\\src\\fmt\\print.go:256:6". After splitting, the file path is "C"! And the line position is incorrect as well.

g2flyer:
Following patch would make go-guru.el work on windows ...
$ diff -u go-guru.el.*
--- go-guru.el.orig     2018-01-22 12:24:52.225719700 -0800
+++ go-guru.el.patched  2018-01-22 12:25:36.845977500 -0800
@@ -324,6 +324,11 @@
   "Find the file containing the position POSN (of the form `file:line:col')
 set the point to it, switching the current buffer."
   (let ((file-line-pos (split-string posn ":")))
+    (if (eq system-type 'windows-nt)
+       (let ((first-file-line-pos (car file-line-pos)))
+         (if (string-match-p "^[a-z]$" first-file-line-pos);; looks like a drive name ...
+             (setq file-line-pos (cdr file-line-pos));; .. so remove it
+           )))
     (funcall (if other-window #'find-file-other-window #'find-file) (car file-line-pos))
     (goto-char (point-min))
     (forward-line (1- (string-to-number (cadr file-line-pos))))

seagle0128:
@g2flyer It looks good. Would you please submit a PR?

g2flyer:
@seagle0128 i just created a pull request (hope it's ok, i have to admit it's my first github pull request :-)

dominikh:
guru is deprecated.

PR #231, opened 2018-02-03
make guru work on windows (iff files are on default drive)

g2flyer:
In response to discussion on issue #218


psanford:
Thanks for the pr! 

The expected format is file:line:col. Instead of checking for windows-nt and something that looks like a widows drive name can we pop the last two values off the list and then join any remaining values on ':'?

dominikh:
guru is unmaintained and doesn't support Go modules. The way forward is to use gopls instead. We won't tinker with the existing implementation.

PR #233, opened 2018-02-09
ignore dep and other non package archive dirs

paulbdavis:
When using [dep](https://github.com/golang/dep), it places a ton of files in the `$GOPATH/pkg` dir, which considerably slows down the time it takes to load the list of installed packages. 

This patch filters out dirs that will not contain package archives before it startes doing the recursive loop.


chessman:
@dominikh Any plans to merge it? I can't use guru anymore because `(go-packages)` hangs on enormous dep cache (`$GOPATH/pkg/dep`).

dominikh:
The entire thing needs to be redesigned on account of the GOCACHE as well as Go modules, anyway. There's not much point in fixing the existing one now.

psanford:
We've deprecated `(go-packages-native)` and switched the default 'go-packages-function to `(go-packages-go-list)` which is much faster for go modules. 

We're not planning on maintaining `(go-packages-native)` going forward so I'm going to close this.

Issue #215, opened 2017-10-04
godoc-gogetdoc breaks on windows


ilohmar:
At least it breaks in my (fairly standard, I think) setup, quoting the drive `:` with double backslashes. 

It seems to me that it is incorrect to call `shell-quote-argument` on the file name (as the above function does), because there is no shell involved at all.  I have removed the quoting call and have it working again on Windows, while it just continues to work on Linux.

ilohmar:
See pull request #235.

PR #235, opened 2018-03-13
rm incorrect filename quoting in godoc-gogetdoc

ilohmar:
This makes it work on Windows as well.  Confer issue #215.


psanford:
Thanks for the pr!

PR #236, opened 2018-03-15
Always move back to same column after applying rcs patch

mschuldt:
Previously the point would move to the beginning of the line if the
current line was changed by gofmt because save-execution stores a text
marker instead of a buffer position.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1029,7 +1029,8 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
         ;; Appending lines decrements the offset (possibly making it
         ;; negative), deleting lines increments it. This order
         ;; simplifies the forward-line invocations.
-        (line-offset 0))
+        (line-offset 0)
+        (column (current-column)))
     (save-excursion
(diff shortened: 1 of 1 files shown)


dominikh:
> save-execution stores a text marker instead of a buffer position

Interesting, that would explain that. I'll give the patch a try soon (though it may take a while, I'm currently swamped, please don't let that discourage you). 

I wonder what the user experience is like, though. Even though point stays in the same position, it may potentially be looking at different syntax. I wonder if that's less confusing than jumping to the first column.

mschuldt:
Personally, most corrections done by gofmt are relatively minor.  If the change is before the point, the point gets shifted a little bit relative to the code, but it's always easy to adjust with a quick char or word movement key if needed. And if the changes are after the point, it does not move.

I suppose in some senses moving to the beginning of line every time is more consistent but from the user perspective it makes little difference, because they are not always on a fixed line when saving and the reason for the fix is not always clear -- most of the time there is no change but then randomly the point moves to the beginning of the line because of some trailing whitespace

PR #238, opened 2018-03-28
Use gopkgs tool to list the packages

cosmincojocar:
This tools is faster than native or go list options.

https://github.com/uudashr/gopkgs

cc  @dominikh


dominikh:
Thank you, but I am not looking to add more external dependencies to go-mode.

PR #240, opened 2018-04-28
Add a simple Makefile to byte-compile the source files

phst:


dominikh:
Is this desirable? Wouldn't it be recommended for people to use `package-install-file` instead?

andreas-roehler:
Being in favor of this PR.  The Emacs packaging has severe issues - when updating some parts, compiled and non-updated stuff might be broken.
Therefor still prefer loading stuff package-by-package from source, not forcibly compiled.

psanford:
I'm going to close this as being out of scope for this project. The few people who want this can do it themselves, either manually the way this pr does it or automatically in their emacs config (see https://www.emacswiki.org/emacs/AutoRecompile for some tips on how to do that).

Issue #46, opened 2014-04-02
godef-jump & godef-describe do not work via TRAMP


ashumilov:
When I use TRAMP 'godef-jump' and 'godef-describe' do not work with message "Could not run godef binary".
When I use them from locally started Emacs they work just fine.

Quasilyte:
I having the same problem.
It seems like the problem is in `call-process-region` as it does not
behaves friendly with remote shell: 
https://lists.gnu.org/archive/html/tramp-devel/2010-02/msg00035.html

Can you re-open this?

dominikh:
Reopening, but don't expect me to work on it. I have no experience with TRAMP internals and no idea how to tackle the issue.

Quasilyte:
I will investigate on that.
Expect a pull request if solution is found. :)
At the very minimum, I will post additional information here for other users that may
encounter this issue in future.

Quasilyte:
Note: because time I can spend on this is limited, it can take a while until issue is solved.

I managed to execute `godef` on remote machine, but other weird issues blocked my way;
to be more precise: `godef` jumped to a wrong position due to incorrectly passed `-o`/offset argument.

[Org-mode own hack](http://orgmode.org/w/?p=org-mode.git;a=commitdiff_plain;h=2056d7d419aabfc9a5fd59674f66e58862f14e32) around `call-process-region` was very useful, 
users may wrap (advice) `call-process-region` to repair it, this requires no changes from the
`go-mode`, but the quality of the result is yet to be discovered.


If I give up with no solution, explicit post will be dropped here.

dominikh:
Thanks for the update and the work on this issue. Take all the time you need. Maybe others will jump in to help, too.

Quasilyte:
Advice for future emacsers: think about tramp alternatives.

There are more than just 2 issues that are outlined in this ticket.
Even if you troubleshoot some of them and make a dirty fix, other things may broke.
More that 1 package uses `call-process-region`, but with fixed version you still will have to
configure many things to make remote environment work.

The solution that is enough to make godef-jump and godef-describe to 
work via tramp is to patch `call-process-region` or advise it (basically, you want to forward all arguments to `org-babel-tramp-handle-call-process-region`).
The second step is optional, but may be required: set cmd location variables (e.g. `godef-command`).
It may be required due to improper tramp remote paths configuration and/or your desire to work with both local and remote Go buffers.

Currently, I gave up this thing. 
[...]

PR #246, opened 2018-05-16
Added TRAMP support in godef--call

shlomiv:
I am not very experienced elisp programmer, so sorry if my attempt is not perfect.. 

I found that by using `process-file` instead of `call-process-region`, godef-jump works for me over tramp connection. Please review and see if this helps others as well.

This should solve #46


shlomiv:
Making a new one..

PR #247, opened 2018-05-16
Added TRAMP support in godef--call

shlomiv:
I am not very experienced elisp programmer, so sorry if my attempt is not perfect..

I found that by using process-file instead of call-process-region, godef-jump works for me over tramp connection. Please review and see if this helps others as well.

This should solve #46

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1489,36 +1489,6 @@ description at POINT."
         (kill-buffer outbuf)
         (split-string result "\n")))))
 
-
-(defun godef--call (point)
-  "Call godef, acquiring definition position and expression
-description at POINT."
-  (if (not (buffer-file-name (go--coverage-origin-buffer)))
-      (error "Cannot use godef on a buffer without a file name")
-    (let ((outbuf (generate-new-buffer "*godef*"))
-          (coding-system-for-read 'utf-8)
-          (coding-system-for-write 'utf-8))
-      (let ((filename (file-truename (buffer-file-name (go--coverage-origin-buffer)))))
-        (if (tramp-tramp-file-p filename)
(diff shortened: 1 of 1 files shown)


cswarth:
This is awesome, nice job.  Had to make a minor edit to adapt the patch to the earlier version of Tramp shipped with Aquamacs, but it worked flawlessly.

go-mode.el
```diff
@@ -1470,24 +1470,26 @@ description at POINT."
...
-                               ;; Emacs point and byte positions are 1-indexed.
-                               (number-to-string (1- (position-bytes point))))
-          (with-current-buffer outbuf
-            (split-string (buffer-substring-no-properties (point-min) (point-max)) "\n"))
-        (kill-buffer outbuf)))))
+          (coding-system-for-write 'utf-8)
+          (filename (file-truename (buffer-file-name (go--coverage-origin-buffer))))
+          (token-position (number-to-string (1- (position-bytes point)))))
+      (if (tramp-tramp-file-p filename)
+          (with-parsed-tramp-file-name filename nil
```
phst: You don't need to use TRAMP functions here.

- Use `file-local-name` to get the local name of a remote file.
- Use `(concat (file-name-remote-p default-directory) filename)` to convert local names on a remote system to remote names.

That way the patch works for all special file names, not just TRAMP.

psanford:
Similar to #207 I don't think it makes sense to add support for TRAMP piecemeal, one command at a time.

I'm hopeful that once gopls is stable this will no longer be an issue as lsp-mode already has support for tramp buffers where the lsp server is running remotely.

dominikh:
LSP and gopls are usable these days, and we won't improve on our support for existing tools like godef et al.

PR #249, opened 2018-05-17
better autocompletion support when seting guru scope

jvshahid:


dominikh:
The guru scope is a lot more free-form than the godoc query. The guru scope can contain the `...` glob character, and may consist of more than one space-separated entry. In the past it also accepted absolute file paths, but I don't know if that is still true. IIRC it also supports negating package selections.

Reusing the godoc input for this is probably not a good fit. I am not sure how viable it would be to write completion that works well for the guru scope. It would have to support a lot of syntax and be context-aware.

jvshahid:
> The guru scope is a lot more free-form than the godoc query. The guru scope can contain the ... glob character, and may consist of more than one space-separated entry. In the past it also accepted absolute file paths, but I don't know if that is still true. IIRC it also supports negating package selections.

That's very useful. I wasn't aware of that feature. I always set the scope to the main package. That said, i still see benefit from having the autocompletion if:

1. you want to set the scope to a single package
1. you want to autocomplete a prefix and append `...`

i don't think we have to overly complicate the completion logic if the user wants to use a more complex scope. WDYT ?

dominikh:
The issue is that the presence of completion often makes it difficult to enter custom text. The default completion will give special meaning to SPC, having it complete entries. And if people use things like ido, ivy, helm etc, it may be even more different.

Thus, the two options are 

1. good completion
2. no completion

Thus, reusing godoc's completion is probably a bad idea.

jvshahid:
I don't think SPC is allowed in the scope according to [the doc](https://docs.google.com/document/d/1_Y9xCEMj5S-7rv2ooHpZNH15JgRT5iM742gJkw5LtmQ/edit). It can only be a comma separated list of packages or a wildcard (i.e. `/...`)

How would ido, ivy or helm interfere with this function, do they replace or advice `completing-read` ?

jvshahid:
I pushed another commit that introduces a new `go-guru--read-scope`. the new function uses `completing-read-multiple` which allows multiple entries to be present in the scope. 

**note** vim set guru scope provides similar behavior to what i'm proposing in this pr. see [here](https://github.com/fatih/vim-go/blob/master/autoload/go/complete.vim#L235) and [here](https://github.com/fatih/vim-go/blob/master/ftplugin/go/commands.vim#L5)

jvshahid:
@dominikh what do you think of the proposed patch ?

dominikh:
Sorry, it'll be a couple days before I can test the patch.

go-guru.el
```diff
@@ -138,6 +138,11 @@
     "---"
     ["Set pointer analysis scope..." go-guru-set-scope t]))
 
+(defun go-guru--read-scope ()
+  "Read a godoc query from the minibuffer."
+  (completing-read-multiple "guru-scope; "
+                            (go-packages) nil nil nil 'go-godoc-history))
```
dominikh: Wrong docstring, and using the wrong history (should be `go-guru--scope-history`). Also, there's a semicolon instead of a colon in the prompt, and the prompt should indicate that multiple, comma-separated elements are permitted. `guru-scope (comma-separated):` might be good enough, unless you can think of something better.

jvshahid:
Fixed the prompt and the history var used.

jvshahid:
I also squashed the two commits to make the history cleaner

dominikh:
Merged at last, thank you for your patience.

jvshahid:
Awesome, thanks!

PR #259, opened 2018-09-14
require subr-x when interpreting go-guru.el

### LINKED ISSUE #258: go-guru.el: go-guru-set-scope depends on subr-x, but does not 'require' it
Whenever I try to evaluate `M-x go-guru-set-scope`, I get the following error:
```
go-guru-set-scope: Symbol’s function definition is void: string-join
```

`M-x version` reports:
```
GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.22.24), modified by Debian
```

From what I can tell, `string-join` comes from the `subr-x` package, and `go-guru-set-scope` succeeds if I explicitly evaluate `(require 'subr-x)` first.

Per the advice in `subr-x`¹, I believe that `go-guru.el` should include the line
```
(eval-when-compile (require 'subr-x))
```

¹https://github.com/emacs-mirror/emacs/blob/9f58ed156974f6345a85da419c57c4235f549f2a/lisp/emacs-lisp/subr-x.el#L33-L34

jvshahid:
Otherwise the user will run into the following error when trying to use
`go-guru-set-scope':

  Symbol’s function definition is void: string-join

This doesn't affect compiled files.

[fixes #258]

### DIFF
--- a/go-guru.el
+++ b/go-guru.el
@@ -143,6 +143,8 @@
   (completing-read-multiple "guru-scope (comma-separated): "
                             (go-packages) nil nil nil 'go-guru--scope-history))
 
+(eval-when-compile (require 'subr-x))
+
 ;;;###autoload
 (defun go-guru-set-scope ()
   "Set the scope for the Go guru, prompting the user to edit the previous scope.


jvshahid:
tested using

```
emacs --batch -L /path/to/go-mode.el/ --eval="(progn (require 'go-guru) (go-guru-set-scope))"
```

konradjniemiec:
Instead of including subr-x, and have string-join still requiring > Emacs 25.3, you could just change `string-join` -> `mapconcat 'identity`

jvshahid:
I don't think that is entirely true. I just tested this with emacs 25.2.2 and it worked.  This made me find out the first version that introduced `string-join` and I believe that to be version `24.4` as evident from the [24 NEWS file](https://github.com/emacs-mirror/emacs/blob/f1ddaf7b65bb8edac41813b48f96a84c7fc2e263/etc/NEWS.24#L1289-L1300).

The question remains, whether 24.4 is a version that we need to support or not. Since commit https://github.com/dominikh/go-mode.el/commit/3139c0f803243d8b2be262b33113641dfe6818e2 tried to fix compatability with Emacs 24.4 I'm inclined to say yes, 24.4 should be supported, but just want to confirm that this is still the case.

jvshahid:
Sorry meant to ask if versions prior to 24 should be supported ?

dominikh:
That commit was back in 2013. I don't think we care about Emacs 23 anymore.

jvshahid:
Cool, so this PR is good to be merged ?

PR #263, opened 2018-11-05
Added go-packages-gopkgs to support gopkgs

nanjj:
`go-packages-native` and `go list all` take seconds when go packages
becoming more. To introduce gopkgs to speed up.


psanford:
I don't think this makes sense to add to go-mode at this point. It is possible for people to accomplish this today by providing the function themselves to `go-packages-function`. 

With gopls on the horizon I don't think it makes sense to add a new function to go-mode that will eventually be deprecated.

PR #264, opened 2018-12-02
Strip vendor package name in go-packages

buzztaiki:
When using [Vendor directories](https://golang.org/cmd/go/#hdr-Vendor_Directories), `go-packages` suggest a package name including a project name like a `github.com/buzztaiki/usedep/vendor/github.com/golang/glog`.

This PR strip preceding paths from vendoring package name in `go-packages`.
I think that it is useful when to use vendoring support tool like a dep.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1391,31 +1391,45 @@ If IGNORE-CASE is non-nil, the comparison is case-insensitive."
 
 
 (defun go-packages ()
-  (funcall go-packages-function))
+  (let ((workspace (go--current-workspace)))
+    (sort
+     (delete-dups
+      (mapcar (lambda (package) (if (not workspace)
+                                    package
+                                  (go--packages-strip-vendor workspace package)))
+              (funcall go-packages-function)))
+     #'string<)))
 
 (defun go-packages-native ()
   "Return a list of all installed Go packages.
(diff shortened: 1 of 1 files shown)


go-mode.el
```diff
@@ -1391,32 +1391,46 @@ If IGNORE-CASE is non-nil, the comparison is case-insensitive."
 
 
 (defun go-packages ()
-  (funcall go-packages-function))
```
psanford: I don't think we want to change the implementation of (go-packages). This allows users to override the default behavior easily by setting the go-packages-function. It would be surprising if we suddenly started manipulating the return results of that based on the current workspace.

go-mode.el
```diff
@@ -2020,6 +2034,10 @@ If BUFFER, return the number of characters in that buffer instead."
   (with-current-buffer (or buffer (current-buffer))
     (1- (position-bytes (point-max)))))
 
+(defun go--current-workspace ()
```
psanford: With go modules I think the name of this function is confusing. If you are outside of the GOPATH your workspace isn't your GOPATH.

I'm also not sure when you would want to match GOROOT. Are you commonly editing files in your GOROOT?

dominikh:
PR stalled, closing.

PR #269, opened 2019-02-01
Speed up godef-jump by avoiding passing -t to godef

lovrop:
I found godef-jump taking more than 10 seconds to find the definition of a private function in a fairly small file (with lots of dependencies though).  This seemed suspicious so I dug into and found the -t flag made godef run much much slower:

% wc -l server.go
     127 server.go
% time godef -f server.go -t getFile > /dev/null
godef -f server.go -t getFile > /dev/null  11.99s user 15.19s system 221% cpu 12.271 total
% time godef -f server.go getFile > /dev/null
godef -f server.go getFile > /dev/null  0.77s user 0.95s system 205% cpu 0.834 total

godef-jump discards the type info anyway so with a bit of code we can avoid requesting it, making the call much faster (in my codebase anyway).

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -255,7 +255,7 @@ doesn't work on package names or variables.
 
 Consider using ‘godoc-gogetdoc’ instead for more accurate results."
   (condition-case nil
-      (let* ((output (godef--call point))
+      (let* ((output (godef--call point t))
              (file (car output))
              (name-parts (split-string (cadr output) " "))
              (first (car name-parts)))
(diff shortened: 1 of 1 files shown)


go-mode.el
```diff
@@ -1474,28 +1474,33 @@ visit FILENAME and go to line LINE and column COLUMN."
       (if (buffer-modified-p)
           (message "Buffer is modified, file position might not have been correct")))))
 
-(defun godef--call (point)
+(defun godef--call (point with-type-info)
```
psanford: I think this change would break external packages depending on (godef--call) as it changes the method signature. A quick search of github shows there are external callers (including company-go).

We probably could do this without breaking other users by using an optional flag, something like `(defun godef--call (point &optional only-location)`.

lovrop: Thanks for taking a look. Makes sense but I'm no longer using this, so no interest in revisiting the PR at this time

PR #205, opened 2017-07-26
Describe gofmt-args option

imsodin:

PR #271, opened 2019-03-30
Fix quotation marks in the gofmt doc string

Miciah:
Fix the quotation marks in the doc string for `gofmt` so that Help mode makes hyperlinks.  Follow-up to https://github.com/dominikh/go-mode.el/pull/205.

Also fix the quotation marks in the doc string for `go--max-dangling-operator-length`.

* `go-mode.el` (`go--max-dangling-operator-length`):
(`gofmt`): Fix quotation marks in doc strings.


psanford:
Thanks for the PR!

PR #279, opened 2019-07-29
Fix call args indent

muirrn:


psanford:
I ran (indent-region (point-min) (point-max)) on a subset of the go stdlib and tool chain to get a sense of what is currently broken and if this change does anything weird. I found one example that is probably worth looking at. On the `go1.12.7` tag src/cmd/compile/internal/gc/esc.go has the following block (gofmt'ed):

```	// Big stuff and non-constant-sized stuff escapes unconditionally.
	// "Big" conditions that were scattered around in walk have been
	// gathered here.
	if n.Esc != EscHeap && n.Type != nil &&
		(n.Type.Width > maxStackVarSize ||
			(n.Op == ONEW || n.Op == OPTRLIT) && n.Type.Elem().Width >= maxImplicitStackVarSize ||
			n.Op == OMAKESLICE && !isSmallMakeSlice(n)) {
		// isSmallMakeSlice returns false for non-constant len/cap.
		// If that's the case, print a more accurate escape reason.
		var msgVerb, escapeMsg string
		if n.Op == OMAKESLICE && (!Isconst(n.Left, CTINT) || !Isconst(n.Right, CTINT)) {
			msgVerb, escapeMsg = "has ", "non-constant size"
		} else {
			msgVerb, escapeMsg = "is ", "too large for stack"
		}

		if Debug['m'] > 2 {
			Warnl(n.Pos, "%v "+msgVerb+escapeMsg, n)
		}
		n.Esc = EscHeap
		addrescapes(n)
		e.escassignSinkWhy(n, n, escapeMsg) // TODO category: tooLarge
	}
```

Without your diff go-mode formats it in the following way (not matching gofmt):
```  	// Big stuff and non-constant-sized stuff escapes unconditionally.
	// "Big" conditions that were scattered around in walk have been
	// gathered here.
	if n.Esc != EscHeap && n.Type != nil &&
		(n.Type.Width > maxStackVarSize ||
		(n.Op == ONEW || n.Op == OPTRLIT) && n.Type.Elem().Width >= maxImplicitStackVarSize ||
		n.Op == OMAKESLICE && !isSmallMakeSlice(n)) {
		// isSmallMakeSlice returns false for non-constant len/cap.
		// If that's the case, print a more accurate escape reason.
		var msgVerb, escapeMsg string
		if n.Op == OMAKESLICE && (!Isconst(n.Left, CTINT) || !Isconst(n.Right, CTINT)) {
			msgVerb, escapeMsg = "has ", "non-constant size"
		} else {
			msgVerb, escapeMsg = "is ", "too large for stack"
		}

		if Debug['m'] > 2 {
			Warnl(n.Pos, "%v "+msgVerb+escapeMsg, n)
		}
		n.Esc = EscHeap
		addrescapes(n)
		e.escassignSinkWhy(n, n, escapeMsg) // TODO category: tooLarge
	}
```

With your change it is indenting differently (still not matching gofmt). I'd say its worse as the body is too far indented and the close curly bracket is no longer aligning with the if statement: 

```  	// Big stuff and non-constant-sized stuff escapes unconditionally.
	// "Big" conditions that were scattered around in walk have been
	// gathered here.
	if n.Esc != EscHeap && n.Type != nil &&
		(n.Type.Width > maxStackVarSize ||
			(n.Op == ONEW || n.Op == OPTRLIT) && n.Type.Elem().Width >= maxImplicitStackVarSize ||
			n.Op == OMAKESLICE && !isSmallMakeSlice(n)) {
			// isSmallMakeSlice returns false for non-constant len/cap.
			// If that's the case, print a more accurate escape reason.
			var msgVerb, escapeMsg string
			if n.Op == OMAKESLICE && (!Isconst(n.Left, CTINT) || !Isconst(n.Right, CTINT)) {
				msgVerb, escapeMsg = "has ", "non-constant size"
			} else {
				msgVerb, escapeMsg = "is ", "too large for stack"
			}

			if Debug['m'] > 2 {
				Warnl(n.Pos, "%v "+msgVerb+escapeMsg, n)
			}
			n.Esc = EscHeap
			addrescapes(n)
			e.escassignSinkWhy(n, n, escapeMsg) // TODO category: tooLarge
		}
```

muirrn:
I think I have fixed the nested dangling situation. Can you run your thing again?

psanford:
For reference, this is what I'm using to apply go mode indenting to a directory recursively: https://gist.github.com/psanford/636cbe569648b8bef85ce22ccedd565b

muirrn:
@psanford can you look at this again? The nested dangling case should be improved now. These are the major remaining indent issues (none of which should be caused by my changes):

- multi-line case statements (will fix in next PR)
- comments above case statements
- dangling comma situations (will fix in next PR)
- complicated dangling boolean expressions involving "||" and "&&" (go fmt indents based on operator precedence)

Performance might be a bit slower but still seems fine. I was going to see if there are any more easy speed ups in a separate PR.

go-mode.el
```diff
@@ -606,31 +644,36 @@ current line will be returned."
...
+             (not (eq (char-after) ?\())
              (go-previous-line-has-dangling-op-p))
-            (- (current-indentation) tab-width)
+            (go--non-dangling-indent)
           (go--indentation-for-opening-parenthesis)))
-       ((progn (go--backward-irrelevant t)
-               (looking-back go-dangling-operators-regexp
-                             (- (point) go--max-dangling-operator-length)))
+       ((let ((dangling-line (go-previous-line-has-dangling-op-p)))
+         (when dangling-line (goto-char dangling-line) (end-of-line) t))
```
psanford: This (let ... (when ... (point movement))) in the cond condition is fairly hard to follow and a bit confusing why we're doing extra work in the condition unrelated to whether it evaluates to true or not. 

Maybe move the let declaration up out of the (cond) and move the movement stuff into the body of this condition?

muirrn: Done.

Issue #282, opened 2019-07-30
Indentation breakage caused by first character open parentheses string literal and apostrophe in comment


psanford:
The following code does not indent properly after the return statement:

```package main

var templTrace = `
(
`

func main() {
	if isUserAnnotation {
		return // don't display unrelated user annotation events.
}
}
```

(This is a minimal example I winnowed down from go/src/cmd/trace/trace.go on the 1.12.7 tag.)

Removing the apostrophe from the comment or not having the open parentheses as the first character on the line in the string literal makes the buffer indent properly.

I've seen some other examples in the stdlib where indentation gets off after a comment with an apostrophe in it. I'll investigate some more to see they are also related to string literals or something else.

dominikh:
Notice the two following things:

1. We're not indenting it incorrectly, we're not indenting it at all for some reason. You could prefix either of the two closing curly braces with any amount of whitespace, and we wouldn't touch it.
2. With show-paren-mode enabled, the two closing curly braces are shown as mismatched. Something about the raw string literal or the comment breaks Emacs's parser.

From what I remember, teaching Emacs about Go's raw string literals required some amount of hackery and custom parsing, because Emacs's own syntax table has no notion of a multi-line string literal that doesn't accept escape sequences. I believe some of that hackery is in the go-propertize-syntax function.

psanford:
This was fixed by #283

PR #283, opened 2019-08-01
Support trailing commas as dangling operators.

### LINKED ISSUE #42: Add comma to go-dangling-operators-regexp ?
before

```
a,b :=
    1,
2
```

after

```
a,b :=
    1,
    2
```

muirrn:
Sometimes a trailing comma behaves like a dangling
operator (i.e. indents following line). For example:

// This comma is a dangling operator since it indents the "2".
return 1,
  2

// This comma isn't a dangling operator since it doesn't indent
// the "2" (the "2" is indented due to the paren block, not the
// dangling comma).
foo(
  1,
  2,
)

// Similarly, not a dangling operator because the indent is due
// to the curlies.
Foo{
  A: 1,
  B: 2,
}

We now have special logic to sometimes treat a trailing comma as a
dangling operator. Specifically, if we are inside parens, a composite
literal or a struct definition, do not treat the comma as a dangling
operator.

Fixes #42

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -66,11 +66,9 @@ Return non-nil if there was a paren to move up to."
   ;; function's outcome.
 
   ;; Silently fail if there's no matching opening parenthesis.
-  (condition-case nil
-      (progn
-        (backward-up-list)
-        t)
-    (scan-error nil)))
+  (let ((open-char (nth 1 (syntax-ppss))))
+    (when open-char
+      (goto-char open-char))))
 
 
 (defconst go-dangling-operators-regexp "[^-]-\\|[^+]\\+\\|[/*&><.=|^]")
--- a/go-mode.el
+++ b/go-mode.el
@@ -58,14 +58,18 @@ function."
                         (progn (forward-visible-line arg) (point))))))
 
 (defun go-goto-opening-parenthesis (&optional _legacy-unused)
-  "Move up one level of parentheses."
+  "Move up one level of parentheses.
+
+Return non-nil if there was a paren to move up to."
   ;; The old implementation of go-goto-opening-parenthesis had an
   ;; optional argument to speed up the function.  It didn't change the
   ;; function's outcome.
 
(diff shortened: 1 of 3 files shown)


go-mode.el
```diff
@@ -616,29 +613,92 @@ current line will be returned."
...
+        ;; preceded by non space (e.g. "Foo|{")
+        (not (looking-back "[[:space:]]" (1- (point))))
+
+        ;; or curly itself is in a composite literal (e.g. "Foo{|{")
+        (go--in-composite-literal-p))))))
+
+(defun go--between-parens-p ()
+  "Return non-nil if point is between '(' and ')'.
+
+The return value is the position of the opening paren."
```
psanford: If the function's return value has meaning beyond true/false then I don't think it should be named *-p (implying its simply a predicate function).

It seems like a better name for this function would be `go--open-paren-position`

psanford:
Functionally this looks good. 

I did notice that there are certain files that doing a full buffer indent runs slower with this change. For example `go/src/cmd/internal/obj/x86/avx_optabs.go` on tag go1.12.7 takes 10s before and 20s after. That file is an outlier but something to be aware of. It looks like files that have very long struct literal definitions are prone to having to do more work to confirm we aren't in the middle of a function call argument list.

Actually editing that file the slow down is not perceivable by me.

go-mode.el
muirrn: Fixed.

muirrn:
I incorporated a change from something else I'm working on to speed up `go-goto-opening-parenthesis`. I couldn't find any change in behavior, but I'm not 100% sure it doesn't break something. @dominikh do you see anything wrong with a77f90d2356073be53a07e17025ff83c696aad1a?

psanford:
This change looks good. The speedup makes the new code faster than before the diff and I don't see any formatting issues running against a subset of the stdlib.

psanford:
The change to `(go-goto-opening-parenthesis)` also fixes #282

PR #284, opened 2019-08-02
Update default for godoc-and-godef-command

### LINKED ISSUE #262: M-x godoc-at-point causes error on latest godoc
I got following error in `*Messages* buffer when run `M-x godoc-at-point`

```
godoc: Unexpected arguments.
usage: godoc -http=localhost:6060
  -analysis string
    	comma-separated list of analyses to perform (supported: type, pointer). See http://golang.org/lib/godoc/analysis/help.html
  -goroot string
    	Go root directory (default "/usr/lib/go")
  -http string
    	HTTP service address (default "localhost:6060")
  -index
    	enable search index
  -index_files string
    	glob pattern specifying index files; if not empty, the index is read from these files in sorted order
  -index_interval duration
    	interval of indexing; 0 for default (5m), negative to only index once at startup
  -index_throttle float
    	index throttle value; 0.0 = no time allocated, 1.0 = full throttle (default 0.75)
  -links
    	link identifiers to their declarations (default true)
  -maxresults int
    	maximum number of full text search results shown (default 10000)
  -notes string
    	regular expression matching note markers to show (default "BUG")
  -play
    	enable playground
  -templates string
    	load templates/JS/CSS from disk in this directory
  -timestamps
    	show timestamps with directory listings
  -url string
    	print HTML for named URL
  -v	verbose mode
  -write_index
    	write index to a file; the file name must be specified with -index_files
  -zip string
    	zip file providing the file system to serve; disabled if empty
```

psanford:
Starting with Go 1.12, the `godoc` command no longer has a
command-line interface (its only a webserver). That
broke (godoc-and-godef) by default.

Originally `godoc` and `go doc` had non-overlapping features which is
why we kept this as `godoc` when we switched 'godoc-command over to
`go doc`. Since that time it looks like the things we need are now
supported in `go doc`, namely you can give `go doc` a filesystem path
instead of a package name.

Fixes #262

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -221,15 +221,13 @@ Set this to nil to upload without prompting.
 
 (defcustom godoc-command "go doc"
   "Which executable to use for `godoc'.
-This can either be 'godoc' or 'go doc', both as an absolute path
-or an executable in PATH."
+This can be either an absolute path or an executable in PATH."
   :type 'string
   :group 'go)
 
-(defcustom godoc-and-godef-command "godoc"
-  "Which executable to use for `godoc' in `godoc-and-godef-command'.
-Must be 'godoc' and not 'go doc' and can be an absolute path or
-an executable in PATH."
+(defcustom godoc-and-godef-command "go doc"
+  "Which executable to use for `godoc-and-godef'.
+This can be either an absolute path or an executable in PATH."
   :type 'string
   :group 'go)
 


go-mode.el
```diff
@@ -226,8 +226,8 @@ or an executable in PATH."
   :type 'string
   :group 'go)
 
-(defcustom godoc-and-godef-command "godoc"
-  "Which executable to use for `godoc' in `godoc-and-godef-command'.
+(defcustom godoc-and-godef-command "go doc"
+  "Which executable to use for `godoc' in `godoc-and-godef'.
 Must be 'godoc' and not 'go doc' and can be an absolute path or
```
muirrn: This doc string doesn't seem right anymore.

psanford: Fixed.

PR #289, opened 2019-08-12
adds a customization for how much of the filename gets truncated

cbbaker:
The length of the filename in the go guru output was hardcoded to 20,
and I wanted it longer, so I added a customization.


psanford:
Can you rebase or squash this PR to get rid of the extra merge commit?

cbbaker:
I think I did this right. Let me know :)

go-guru.el
```diff
@@ -231,9 +236,10 @@ output of the Go guru tool."
 	  (setq p (1- p)) ; exclude final space
 	  (let* ((posn (buffer-substring-no-properties start p))
 		 (flen (cl-search ":" posn)) ; length of filename
-		 (filename (if (< flen 19)
+		 (truncate (- (max go-guru-truncate-file-length 20) 1))
```
dominikh, line 239: I don't understand the use of `max` here. Can we not truncate to less than 20 characters?

PR #290, opened 2019-08-14
Fix complicated dangling indent

muirrn:


go-mode.el
```diff
@@ -719,42 +644,198 @@ The return value is the position of the opening paren."
        (point)))))
 
 (defun go-indentation-at-point ()
+  "Return the appropriate indentation for the current line.
+
+This function works by walking a line's characters backwards. When it
+encounters a closing paren or brace it bounces to the corresponding
+opener. If it arrives at the beginning of the first line, it moves to
```
psanford: 'First line' is a bit confusing/ambiguous. I think you mean 'the line you are indenting'.

psanford:
This cuts reindent time by more than half. Nice work.

go-mode.el
muirrn: Done

PR #292, opened 2019-08-15
indent: fix nested dangling lines that don't use parens

muirrn:
We weren't handling cases like:

1. if foo ||
2.   foo &&
3.    foo {
4.   code()
5. }

We were indenting line 4 too far because we weren't following the
dangling lines all the way up to line 1. There can be arbitrary
amounts of nested indents with no parentheses, so to find the indent
inside the "if" block you must follow all dangling lines to the
beginning of the statement.

I also fixed a couple other minor indent issues:
 - fix an edge case with multi-line case statements where we weren't
   indenting the final line with the colon
 - fix go--end-of-line to skip multiple /* */ style comments at the
   end of the line

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -590,6 +590,7 @@ case keyword. It returns nil for the case line itself."
                 (or
                  (go--boring-line-p)
                  (go--line-suffix-p ","))
+                (not (looking-at go--case-regexp))
                 (go--forward-line -1)))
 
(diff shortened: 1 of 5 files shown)


psanford:
I've been working on some automation for testing pull requests. Lets give it a shot for this pr.

@go-mode-bot run

PR #293, opened 2019-08-18
indent: fix case statement in func in composite literal

muirrn:
Our composite literal detection wasn't working in cases like:

Foo{
  Bar: func() {
    switch {
      case qux: // indents too far here

We thought the "func" open curly was a composite literal curly with
implicit type. Fix by tweaking the logic to instead check that the
open curly is preceded by nothing, a comma, or a colon.


muirrn:
@go-mode-bot run

Issue #119, opened 2016-02-04
go-mode should set `adaptive-fill-regexp`


timcharper:
Presently, `adaptive-fill-regexp` is not set, causing `fill-region` to not reformat comments properly.

I believe the regexp from `c-mode` can be safely borrowed (since commenting `//` and `/*` are the same). But there's probably some special comment syntax in one lang or the other of which I'm not aware.

Anyways, I borrowed the one from `c-mode` and it works fine. I call this in a function included in `go-mode-hook`.

```
(set (make-local-variable 'adaptive-fill-regexp)
       "[   ]*\\(//+\\|\\**\\)[     ]*\\([  ]*\\([-–!|#%;>*·•‣⁃◦]+[  ]*\\)*\\)")
```

dominikh:
Is this related to #91? Also, is there a solution that doesn't require a quite as horrifying regular expression? Maybe it can be stripped down? At a glance, it looks like it handles a lot more special formatting than we really need.

dominikh:
Also, when you say `c-mode`, do you mean `cc-mode`? Because in `cc-mode`, I can't find where it'd be setting that variable, and `cc-mode` uses its own function for filling (`c-fill-paragraph`)

timcharper:
Horrifying reg ex is the only option of which in aware. I pulled this one from c-mode

codeasone:
That really is a horrific reg ex but it's working for me. Thanks @timcharper.

codeasone:
Actually, I spoke too soon. It fails for the following 

```
////////////////////////////////////////////////////////////////////////////////
// Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
////////////////////////////////////////////////////////////////////////////////
``` 

After `M-x fill-paragraph` with the above selected, I'm getting: 

```
////////////////////////////////////////////////////////////////////////////////
// Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod //
tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
////////////////////////////////////////////////////////////////////////////////
``` 

But still, the work-around you suggested @timcharper helps in a number of other scenarios.

codeasone:
[...]

PR #294, opened 2019-08-20
Improve fill-paragraph for comment blocks.

### LINKED ISSUE #91: M-Q adds extra /* in /* */ style comments
I am using Emacs 2.45, and the most recent version of go-mode.el. In Go mode, M-Q sometimes adds an extra /*

<pre>
/* isgb returns true if the decomposition is valid for GB (simplified Chinese).

This has a couple of issues. */
</pre>

Put the pointer on the first /\* without selecting a region, and It comes out like this;

<pre>
/* isgb returns true if the decomposition is valid for GB (simplified
/* Chinese).

This has a couple of issues. */
</pre>

If the whole paragraph is selected first and then M-Q is pressed, you get a different result:

<pre>
/* isgb returns true if the decomposition is valid for GB (simplified
Chinese).

This has a couple of issues. */
</pre>

Is it possible to co-opt c-fill-paragraph from lisp/progmodes/cc-cmds.el in the emacs distribution for these kinds of comments?

Incidentally I also tested this is not a setup bug by running 

```
 emacs -Q gbdecomp.go
```

and then importing go-mode.el and using eval-buffer, then switching on go-mode in the gbdecomp.go buffer using M-x go-mode. The comment indenting was the same.

muirrn:
Implement a few choice functions so that fill-paragraph and friends
works better for Go comment blocks. The two keys are:

- go--fill-prefix which extracts the fill prefix for the following
  comment block. This knows, for example, that "/*" is not a prefix
  that needs to be applied to every comment.
- go--fill-forward-paragraph which essentially defines what a comment
  paragraph is. fill-paragraph works by jumping forward
  with (fill-forward-paragraph 1) and then jumping back
  with (fill-forward-paragraph -1). That defines the start and end of
  the region to be filled.

Fixes #91, #119

@go-mode-bot run

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -632,6 +632,106 @@ case keyword. It returns nil for the case line itself."
        ;; line or the previous character is a comma or colon.
        (or (bolp) (looking-back "[,:]" (1- (point)))))))))
 
+(defun go--fill-prefix ()
+  "Return fill prefix for following comment paragraph."
+  (save-excursion
+    (beginning-of-line)
+
+    ;; Skip over empty lines and empty comment openers/closers.
+    (while (and
+            (or (go--empty-line-p) (go--boring-comment-p))
+            (zerop (forward-line 1))))
+
+    ;; If we are looking at the start of an interesting comment, our
+    ;; prefix is the comment opener and any space following.
+    (if (looking-at (concat go--comment-start-regexp "[[:space:]]*"))
+        ;; Replace "/*" opener with spaces so following lines don't
+        ;; get "/*" prefix.
+        (replace-regexp-in-string "/\\*" "  "
+                                  (match-string-no-properties 0)))))
+
+(defun go--fill-paragraph (&rest args)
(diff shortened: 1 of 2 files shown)


muirrn:
@go-mode-bot run

go-mode.el
```diff
@@ -632,6 +632,94 @@ case keyword. It returns nil for the case line itself."
...
+(defun go--fill-paragraph (&rest args)
+  "Wrap fill-paragraph to set custom fill-prefix."
+  (let ((fill-prefix (go--fill-prefix))
+        (fill-paragraph-function nil))
+    (apply 'fill-paragraph args)))
+
+(defun go--empty-line-p ()
+  (looking-at "[[:space:]]*$"))
+
+(defun go--boring-comment-p ()
```
psanford: Can you add a comment explaining what a boring comment is?

Also its a little confusing that there is both a boring-comment-p and an interesting-comment-p function. One might reasonably expect those to be mutually exclusive when executed from within a comment.

PR #295, opened 2019-08-22
make it possible for godoc-at-point to reuse a single buffer for all lookups

### LINKED ISSUE #180: reuse a single *godoc* buffer for godoc-at-point
Currently, `godoc-at-point` opens a separate buffer for each call. Such a buffer is named something similar to `*godoc /path/to/ Type*`. I always seem to forget to close these buffers, so they quickly pile up. 

I would much prefer having all these `godoc-at-point` calls displaying in a single buffer, say `*godoc*`, whose contents would be replaced for every call. If the old behavior is preferred (one separate buffer per call), I would love to see an option. For example

    ;; reuse a single *godoc* buffer to display godoc-at-point calls
    (setq go-godoc-reuse-buffer t)  

Would this make sense?

As a side-note, these buffers seem to also cause problems for me with the neotree project explorer open. Successive calls to `godoc-at-point` opens/splits buffers in a somewhat irrational way (the `*godoc*` buffers, which I am used to seeing at the bottom of the screen, wander around and all of a sudden `*scratch*` and `*Messages*` appear). I suspect that this is due to some weirdness in neotree, but it works fine for me, for example, in `python-mode` and I suspect that it would work better in `go-mode` if only a single `*godoc*` buffer was being reused.

petergardfjall:
This introduces a customization variable `godoc-reuse-buffer` which, when
non-`nil`, makes `godoc-at-point` reuse the same (`*godoc*`) buffer to write output
from all invocations. The default behavior (when `nil`), is still to write each
output to a separate buffer (`*godoc <path>*`).

This resolves #180.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -239,6 +239,12 @@ Only really desirable when using `godoc' instead of `go doc'."
   :type 'boolean
   :group 'godoc)
 
+(defcustom godoc-reuse-buffer nil
+  "Reuse a single *godoc* buffer to display godoc-at-point calls.
+The default behavior is to open a separate buffer for each call."
+  :type 'boolean
+  :group 'godoc)
+
 (defcustom godoc-at-point-function #'godoc-and-godef
   "Function to call to display the documentation for an
 identifier at a given position.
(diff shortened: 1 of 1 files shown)


psanford:
Thanks for the pr!

PR #299, opened 2019-09-10
update default go-packages-function to use go list

### LINKED ISSUE #298: Consider changing the go-packages-function default to 'go-packages-go-list (and what to do about go-package-native)
The current default for 'go-packages-function is 'go-packages-native. Based on the variable documentation, I think the rationale for that was that go-packages-go-list was historically slower than go-packages-native for many environments. 

Modules seem to change the equation. At least in my environment when I'm working in a module project, go-packages-native is painfully slow compared to go-packages-go-list. I think part of this is there is a lot more stuff that the go tool has been sticking in $GOPATH/pkg in recent releases that we aren't currently skipping. There can now also be multiple versions of packages in there due to modules. 

I also think that `go list` is probably faster for modules than it used to be for GOPATH packages as now it can cull packages that are not in go.mod dependencies. 

Another issue with go-packages-native is it doesn't understand the module pathnames under pkg. So for example it includes the following incorrect package name for me right now:  

```
golang.org/x/tools@v0.0.0-20190213223815-88f95592de8c/go/gccgoexportdata/testdata/long
```

I think it makes sense to switch the default 'go-packages-function to 'go-packages-go-list and to mark go-packages-native as deprecated.

psanford:
The previous default, go-packages-native doesn't work well with the Go
build cache and Go modules. Fixing it would require us to reimplement
a lot of the `go list` logic in elisp (and continue to maintain it for
future changes to the go tool). It doesn't make sense to do that as
the original reason for the elisp implementation (speed) is much less
of an issue when using modules.

Fixes #298

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -185,15 +185,15 @@ a `before-save-hook'."
   :type '(repeat (list regexp (choice (repeat string) function)))
   :group 'go)
 
-(defcustom go-packages-function 'go-packages-native
+(defcustom go-packages-function 'go-packages-go-list
   "Function called by `go-packages' to determine the list of available packages.
 This is used in e.g. tab completion in `go-import-add'.
 
-This package provides two functions: `go-packages-native' uses
+This package provides two functions: `go-packages-go-list' uses
+'go list all' to determine all Go packages. `go-packages-native' uses
 elisp to find all .a files in all /pkg/ directories.
-`go-packages-go-list' uses 'go list all' to determine all Go
-packages.  `go-packages-go-list' generally produces more accurate
-results, but can be slower than `go-packages-native'."
+`go-packages-native' is obsolete as it doesn't behave correctly with
+the Go build cache or Go modules."
   :type 'function
   :package-version '(go-mode . 1.4.0)
   :group 'go)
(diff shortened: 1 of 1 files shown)


go-mode.el
```diff
@@ -1950,8 +1950,11 @@ If IGNORE-CASE is non-nil, the comparison is case-insensitive."
   (funcall go-packages-function))
 
 (defun go-packages-native ()
-  "Return a list of all installed Go packages.
-It looks for archive files in /pkg/."
+  "Return a list of all installed Go packages. Obsolete.
```
muirrn: Did you consider declaring this function as [obsolete](https://www.gnu.org/software/emacs/manual/html_node/elisp/Obsolete-Functions.html)?

go-mode.el
```diff
@@ -1950,8 +1950,11 @@ If IGNORE-CASE is non-nil, the comparison is case-insensitive."
...
+It looks for archive files in /pkg/. This is strategy does not work
```
muirrn: > This is strategy

Typo

psanford: Fixed

go-mode.el
psanford: I did consider it. The reason I didn't do it was I didn't want to trigger the warning message (yet). Maybe I'm being too conservative but it seem like it would be fine to just switch the default and update the docs. I was thinking we would wait until a majority of projects have switched to modules and then use `(define-obsolete-function-alias)` to remove this function.

PR #302, opened 2019-09-22
Add major mode for go.mod files: go-dot-mod-mode

psanford:
This is a simple major mode for go.mod files. It's named
go-dot-mod-mode instead of go-mod-mode to avoid tab completion
conflicts with go-mode* functions.

- It font locks key words and comments. 
- I just borrowed the existing go-mode indentation code since it already does the right thing for go.mod files.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -2635,11 +2635,11 @@ If BUFFER, return the number of characters in that buffer instead."
 
 (defconst go-dot-mod-mode-keywords
   '("module" "go" "require" "replace" "exclude")
-  "All keywords in the Go language.  Used for font locking.")
+  "All keywords for go.mod files.  Used for font locking.")
 
 (defvar go-dot-mod-font-lock-keywords
   `(
-    (,(concat "\\_<" (regexp-opt go-dot-mod-mode-keywords t) "\\_>") . font-lock-keyword-face))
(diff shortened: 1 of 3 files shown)


go-mode.el
```diff
@@ -2579,6 +2579,50 @@ If BUFFER, return the number of characters in that buffer instead."
...
+  (let ((st (make-syntax-table)))
+    ;; handle '//' comment syntax
+    (modify-syntax-entry ?/ ". 124b" st)
+    (modify-syntax-entry ?\n "> b" st)
+    st)
+  "Syntax table for `go-dot-mod-mode'.")
+
+(defconst go-dot-mod-mode-keywords
+  '("module" "go" "require" "replace" "exclude")
+  "All keywords in the Go language.  Used for font locking.")
```
muirrn: > All keywords in the Go language

Copypasta?

psanford: fixed.

psanford:
I updated the pr to better handle the issue you found around keywords in urls.

muirrn:
Looks great, good work!

I feel like some fontification for the module name and/or versions might help with readability. For example, give the module name `font-lock-type-face`. 

Screen Shot 2019-10-16 at 11 40 53 AM

Screen Shot 2019-10-16 at 11 43 46 AM

Screen Shot 2019-10-16 at 11 46 34 AM

Screen Shot 2019-10-16 at 11 47 28 AM

PR #305, opened 2019-10-03
Add Interactive Go Commands

kcotugno:
Just adds a few interactive functions to simply running common Go commands.

- `go-build` 
- `go-test`
- `go-vet`


psanford:
Thanks for the pr. Considering how trivial these functions are and that we haven't already added them to go-mode I don't think we're going to add these now.

PR #308, opened 2019-10-15
Run all tests in GH CI

muirrn:
And fix go-fill-paragraph-test.el to pass from the command line. Not
sure exactly what is wrong, but fill-paragraph wasn't doing the right
thing with a region selected when invoked from the command line. There
could be a real bug here, but I wasn't able to reproduce the failures
functionally in Emacs.

@go-mode-bot run

### DIFF
--- a/.github/workflows/emacs.yml
+++ b/.github/workflows/emacs.yml
@@ -14,4 +14,4 @@ jobs:
       uses: actions/checkout@v1
 
     - name: Test
-      run: cd test; emacs --batch -q -l ert -l ../go-mode.el -l go-indentation-test.el -f ert-run-tests-batch-and-exit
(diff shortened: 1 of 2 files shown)


.github/workflows/emacs.yml
```diff
@@ -14,4 +14,4 @@ jobs:
       uses: actions/checkout@v1
 
     - name: Test
-      run: cd test; emacs --batch -q -l ert -l ../go-mode.el -l go-indentation-test.el -f ert-run-tests-batch-and-exit
+      run: cd test; emacs --batch -q -l ert -l ../go-mode.el $(for t in `ls *-test.el`; do echo -n "-l $t "; done) -f ert-run-tests-batch-and-exit
```
psanford: You don't need to invoke for this `ls`: `for t in *-test.el`.

muirrn: Done

PR #311, opened 2019-10-18
Fontify types

muirrn:


psanford:
Looks good. Here's screenshots of the fontification changes for reference:

Type switches:
type-switch-before
type-switch-after

Composite literals:
composite-literal-before
composite-literal-after

Array with named size:
array-const-before
array-const-after

PR #312, opened 2019-10-18
Improve imenu regex for multi-line func signatures

### LINKED ISSUE #57: imenu-generic-expression does not match all function styles

psanford:
Make the imenu func regex handle function signatures split across
multiple lines.

Fixes #57

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -1670,7 +1670,7 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
 
   (setq imenu-generic-expression
         '(("type" "^type *\\([^ \t\n\r\f]*\\)" 1)
-          ("func" "^func *\\(.*\\) {" 1)))
+          ("func" "^func *\\([^{]*\\) {" 1)))
   (imenu-add-to-menubar "Index")
 
   ;; Go style


go-mode.el
```diff
@@ -1670,7 +1670,7 @@ with goflymake \(see URL `https://github.com/dougm/goflymake'), gocode
 
   (setq imenu-generic-expression
         '(("type" "^type *\\([^ \t\n\r\f]*\\)" 1)
-          ("func" "^func *\\(.*\\) {" 1)))
+          ("func" "^func *\\([^{]*\\) {" 1)))
```
muirrn: Why do we have to match beyond the function name? Does imenu do something with the args or return values?

psanford: I menu just displays what you match. I think the original regex was written that way to properly match both package level functions and methods on structs. I tried to maintain what we were already matching while supporting a split across newlines.

muirrn: If that is the case then your new regex breaks on signatures like:
```go
func foo(s struct { i int }, i interface { bar() }) {
```

But, based on my testing, it isn't the case. It seems like the regex only needs to match the function name.

8i8:
The following regex works fine:

"^func\s\\(.*$\\)"

It does not display the extra lines when the function call is multi line, it dispalays the first line in its entirety.

dominikh:
PR stalled, closing.

PR #321, opened 2019-11-18
Fix Content-Type for https://play.golang.org/share

xuchunyang:
I pushed another unrelated commit because it's just a typo.


psanford:
Looks good. Thanks for the pr!

PR #329, opened 2020-01-04
godoc-and-godef understands packages

sfllaw:
godoc-and-godef can now parse the package name from the import statements returned by godef. Go doc is then called to display the package's documentation.

Closes: #323

Issue #323, opened 2019-12-02
godoc-at-point fails on package names


sfllaw:
With `GO111MODULE=on`, `godoc-at-point` is unable to resolve `os`:
``` go
package main

import "os"

func main() {
	os.Exit(0)
//	^ call godoc-at-point here
}
```

with the following message: `godoc: doc: invalid identifier "import"`

I believe this is because `(godef--call (point))` returns the following message:
```
("main.go:6:2" "import (os \"os\")" "")
```

This is documented as being unhandled by `godoc-and-godef`, but it should be relatively easy to change if the following code was able to parse the `import` statement for the third result, _i.e._ the full package name:
https://github.com/dominikh/go-mode.el/blob/cad6d0a522f8f5c351a1332bcd54bd5ffd8ce30d/go-mode.el#L276-L278

sfllaw:
Here’s a version that works for me:
``` elisp
(defun godoc-and-godef (point)
  "Use a combination of godef and godoc to guess the documentation at POINT.

Due to a limitation in godoc, it is not possible to differentiate
between functions and methods, which may cause `godoc-at-point'
to display more documentation than desired.  Furthermore, it
doesn't work on package names or variables.

Consider using ‘godoc-gogetdoc’ instead for more accurate results."
  (condition-case nil
      (let* ((output (godef--call point))
             (file (car output))
             (name-parts (split-string (cadr output) " "))
             (first (car name-parts)))
        (if (not (godef--successful-p file))
            (message "%s" (godef--error file))
          (go--godoc
           (if (string= first "import")
               (let ((pkg-name (string-remove-suffix ")" (caddr name-parts))))
                 (car (split-string-and-unquote)))
             (format "%s %s"
                     (file-name-directory file)
                     (if (or (string= first "type") (string= first "const"))
                         (cadr name-parts)
                       (car name-parts))))
           godoc-and-godef-command)))
    (file-error (message "Could not run godef binary"))))```

psanford:
Can you send a PR for your fix?

sfllaw:
@psanford Any chance you could look at #329, #330, and #331?

sfllaw:
@psanford Just a friendly ping to look at #329, #330, and #331 please?

sfllaw:
@psanford Ping? Please review #329, #330, and #331?

sfllaw:
@psanford I have sent in a fix as requested, is there something you would like me to revise?

sfllaw:
[...]

PR #330, opened 2020-01-04
godoc-and-godef for unexported symbols

sfllaw:
godoc-and-godef can now show the documentation for unexported symbols, much like gogetdoc.

Go doc will display the documentation for an unexported symbol only when the point is on the symbol itself: it will not display unexported symbols inside packages or structs, as that would be too noisy.

Requires: #329


sfllaw:
@dominikh @muirdm @psanford: Any chance I could get this reviewed, please?

sfllaw:
Withdrawn because gopls will eventually replace this functionality. See https://github.com/dominikh/go-mode.el/issues/323#issuecomment-967795982

PR #331, opened 2020-01-04
godoc-gogetdoc works with modules

sfllaw:
godoc-gogetdoc now works with modules outside GOPATH by creating a temporary buffer using with-temp-buffer, which preserves default-directory.


sfllaw:
@dominikh @muirdm @psanford: Any chance I could get this reviewed, please?

PR #351, opened 2020-04-03
go-mode: Improve successful output parsing to detect newest godef

wuurrd:
giving output: no object


psanford:
I've merged this. Thanks!

PR #353, opened 2020-04-10
Allow leading spaces for compilation error message

yuezhu:
In compilation buffer, the error message from `go test` containing the file and line number may begin with spaces. My go version: `go1.14.1 darwin/amd64`.


psanford:
Thanks, this looks correct. 

I've been using a custom compilation mode so I didn't notice this was broken.

Issue #362, opened 2020-05-10
go-test compilation error pattern doesn't match Go 1.14 output


phst:
The output of `go test -v` has changed somewhat in Go 1.14.  For example, with this `/tmp/f/foo_test.go`:

```go
package foo_test

import "testing"

func Test(t *testing.T) {
	t.Error(1)
}
```

the Go 1.13 output is

```
=== RUN   Test
--- FAIL: Test (0.00s)
    foo_test.go:6: 1
FAIL
exit status 1
FAIL	_/tmp/f	0.010s
```

but the Go 1.14 output is

```
=== RUN   Test
    Test: foo_test.go:6: 1
--- FAIL: Test (0.00s)
FAIL
exit status 1
FAIL	_/tmp/f	0.011s
```

Note that the error message now has the prefix `Test: ` (i.e. test name, colon, space). This is probably related to the change described at https://golang.org/doc/go1.14#go-test. However, the `go-test` pattern treats the entire `Test: foo_test.go` part as filename.

`go-test` should be adapted accordingly.

dominikh:
I cannot reproduce this with go1.18-8ff254e30b:

```
=== RUN   Test
    sand_test.go:6: 1
--- FAIL: Test (0.00s)
FAIL
exit status 1
FAIL	example.com/sand2	0.001s
```

phst:
Maybe this got fixed again by a change in Go 1.15 (https://go.dev/doc/go1.15#testing):

> `go test -v` now groups output by test name, rather than printing the test name on each line.

It's probably still worth adding a unit test to ensure that Emacs correctly parses the test output.

PR #363, opened 2020-05-10
Fix ‘go-test’ compilation message pattern.

### LINKED ISSUE #361: Regular expression for go-test compilation pattern is too broad
The regular expression to find Go test errors is currently (https://github.com/dominikh/go-mode.el/blob/734d5232455ffde088021ea5908849ac570e890f/go-mode.el#L1845) `"^\\s-+\\([^()\t\n]+\\):\\([0-9]+\\):? .*$"`. The problem is that this regex is very broad and matches too much:

1. The `\s-+` also matches newlines (at least when using the standard syntax table).
2. The `[^()\t\n]+` group matches a ton of things: spaces, exotic characters, NUL bytes, ...

Combined, these two mean that e.g. a GNU-style `file:line:column: message` line that follows a newline is matched incorrectly:

```lisp
(with-syntax-table (standard-syntax-table)
  (let ((regex "^\\s-+\\([^()\t\n]+\\):\\([0-9]+\\):? .*$")
        (line "\nfile.go:1:2: word word"))
    (list
     (string-match regex line)
     (match-string 1 line))))

⇒ (0 "file.go:1")
```

Here, this incorrectly treats `file.go:1` as filename and 2 as line number.
This causes issues because such lines are very common in practice, and the leading newline alone is enough to trigger `go-test` matching.

I recommend restricting the `go-test` pattern:

1. Replace the `\s-+` by a more precise match, e.g. `    `.
2. Replace the filename group by something that is likely to be an actual filename (e.g. no spaces or colons).

phst:
Fixes #361, #362

Restrict the prefix and filename pattern (Bug #361), and also allow for a test
name prefix (Bug #362).

Add regression tests for these bugs.


test/go-compile-test.el
```diff
@@ -0,0 +1,49 @@
+;;; go-compile-test.el --- unit tests for ‘go-mode’ compilation -*- lexical-binding: t; -*-
+
+;; Copyright 2020 Google LLC
```
psanford, line 3: Is there a reason you used a different copyright assignment and license for this file than normal go-mode.el?

phst: I thought this would be a requirement for Googlers to contribute patches, but it's not (https://opensource.google/docs/patching/#license-headers-and-copyright-notices), so I've changed the copyright notice to "The go-mode authors".

PR #366, opened 2020-06-30
fix: go-guru-definition jump to wrong position by wrong coding-system.

tangxinfa:
When default coding-system is "utf-8-unix" and open a file with "utf-8-dos"
coding-system, go-guru-definition will jump to wrong position.

This is because windows's EOL is "\r\n", is different with linux's "\n", so we
must keep the EOL unchanged when process internally.


tangxinfa:
go-guru-hl-identifier also has the similar problem, but i don't know how to fix it.

dominikh:
guru is unmaintained and doesn't support Go modules. The way forward is to use gopls instead. We won't tinker with the existing implementation.

PR #383, opened 2021-02-22
Fix fontification of multiplicand after index expr.

### LINKED ISSUE #382: Bad font locking for multiplication
In

```go
func _() {
	foo[1]*10 + 1
}
```
we mistakenly color "10" as a type.

muirdm:
In "foo[1]*10" we were interpreting "[1]*10" as an array with element
type "*10". At first I was going to fix the identifier regex to
disallow leading numbers (so "10" couldn't be a type name), but that
is involved and doesn't fix the similar case of "foo[1]*foo[2]".
Instead fix the array type detection to disallow an identifier
character preceding the opening bracket.

Fixes #382.

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -500,7 +500,7 @@ statements."
      (,(concat go-type-name-regexp "{") 1 font-lock-type-face)
 
      ;; Map value type
-     (,(concat "\\_<map\\_>\\[[^]]+\\]" go-type-name-regexp) 1 font-lock-type-face)
+     (go--match-map-value 1 font-lock-type-face)
 
      ;; Map key type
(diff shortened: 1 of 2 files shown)


dominikh:
Does that mean `var x chan[5]int` will not highlight correctly, now?

dominikh:
I guess that's acceptable, since that's never gofmt'ed. I missed that `foo[1]*10` can actually occur in a gofmt'ed context.

muirdm:
This breaks fontification of "float64" in `map[string]map[string]float64`. It appears we were depending `[string]float64` looking like an array.

muirdm:
Closing for now.

muirdm:
OK - I think I resolved the "map" issues I noted above.

PR #399, opened 2022-01-14
Fix two byte-compiler warnings

skangas:
This fixes two byte-compiler warnings in Emacs 28:
```
In go--in-function-p:
go-mode.el:2627:8: Warning: docstring wider than 80 characters

In go-set-project:
go-mode.el:2798:8: Warning: docstring wider than 80 characters
```


dominikh:
Thanks.

Issue #393, opened 2021-12-14
Add support for type parameters


dominikh:

muirdm:
I was looking at this, at least in terms of font locking. For ambiguous cases like `foo[bar](baz)` do you think we should assume it is a generic function instantiation and treat "bar" as a type, or leave it as-is?

dominikh:
I say we leave it as is.

muirdm:
Beyond #400, the only other thing I'm aware of for fontification is fontifying function calls and composite literal type names properly in `foo[int](123)` and `some.Struct[int]{foo: bar}`. Anything else?

dominikh:
I'm the wrong person to ask about fontification, I barely use any. But I'd like to err on the side of "no fontification" over "wrong fontification". So since we lack type info, we can't do anything for `foo[int](123)`.

PR #400, opened 2022-03-26
Generic fontification

muirdm:
Attempt to fix some fortification issues with type params

#393

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -453,6 +453,16 @@ statements."
       (1 font-lock-type-face nil t)
       (2 font-lock-type-face nil t))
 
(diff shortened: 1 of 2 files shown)


dominikh (approved):

dominikh:
@muirdm is this ready for merging as is?

muirdm:
> is this ready for merging

No, I'm not really happy with this. We can peel off just the first commit for now for some relief.

I have a different branch with a somewhat better approach - let me try to get that ready for eyes.

muirdm:
I opened #410 that has a better approach where we can "piggyback" on top of the existing type fontification in most cases.

PR #403, opened 2022-04-17
Remove go--fill-prefix

PuercoPop:
The function is not called from anywhere and does not work as it calls
function that were previously deleted. Both go--empty-line-p and
go--boring-comment-p were removed in
b212a73c21a8b806f9764004397c1d03eb0b2e10.


mediogre (approved):
This fixes comp warnings:
- Warning: the function ‘go--boring-comment-p’ is not known to be defined
- Warning: the function ‘go--empty-line-p’ is not known to be defined.

dominikh:
Thanks!

PR #408, opened 2022-07-26
go-dot-work-mode

wyuenho:
Keywords are taken from the grammar [specified here](https://go.dev/ref/mod#workspaces).

Thanks for this wonderful package!


dominikh:
Thanks!

PR #419, opened 2023-06-24
builtins: add new go 1.21 builtins

jrockway:
This adds `min`, `max`, and `clear`, as per: https://tip.golang.org/doc/go1.21#language


dominikh:
Thank you.

PR #420, opened 2023-07-21
Fix obsoleted function warnings for newer Emacs

syohex:
There is the following byte-compile warning on Emacs 29 or higher versions. This patch fixes it.

```
In go-guru--goto-byte-column:
go-guru.el:321:52: Warning: ‘point-at-bol’ is an obsolete function (as of
    29.1); use ‘line-beginning-position’ or ‘pos-bol’ instead.
```


dominikh:
Deprecating an alias. Great.

PR #433, opened 2024-01-17
Make `indent-tabs-mode' buffer local

shynur:


dominikh:
I don't believe this is necessary? According to its documentation, `indent-tabs-mode` automatically becomes buffer-local when set.

shynur:
Oops, sorry!
I indeed overlooked this.
Due to my configuration issues, the value of `indent-tabs-mode` was never what I expected.
Modifying the code in `go-mode` _seemed_ to have solved my problem, but now it looks like I must have missed something.
How embarrassing.

This can be closed now.
(Yet I think it's not bad to use `setq-local` to explicitly indicate the variable is buffer-local.)

dominikh:
> Yet I think it's not bad to use setq-local to explicitly indicate the variable is buffer-local

I agree, but we'll do that as part of the tree-sitter rewrite of go-mode. Thanks!

PR #439, opened 2024-06-10
Add go-{browse-{freesymbols,doc,assembly},rename}

adonovan:
These four helper functions provide convenient ways to access gopls server functionality without the user needing to worry about how their LSP client does things. They should work for eglot and lsp-mode (though I haven't tested the latter).

Updates dominikh/go-mode.el/#436

### DIFF
--- a/go-mode.el
+++ b/go-mode.el
@@ -3051,6 +3051,45 @@ This handles multi-line comments with a * prefix on each line."
 
 
 
+;; Convenient go-* functions for gopls features available through code
+;; actions, that work across LSP clients:
+
+(defun go-mode--code-action (kind)
+  "Request and invoke the specified kind of code actions for the current selection."
+  (cond
+   ((and (boundp 'eglot--managed-mode) eglot--managed-mode)
+    (let ((beg-end (eglot--code-action-bounds)))
+      (eglot-code-actions (car beg-end) (cadr beg-end) kind t)))
+   ((and (boundp 'lsp-mode) lsp-mode)
+    (lsp-execute-code-action-by-kind kind))
+   (error "buffer is not managed by a known LSP client")))
+
+(defun go-browse-freesymbols ()
+  "View free symbols referenced by the current selection in a browser. Requires gopls v0.16."
+  (interactive)
+  (go-mode--code-action "source.freesymbols"))
+
+(defun go-browse-doc ()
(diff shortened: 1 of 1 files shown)


go-mode.el
```diff
@@ -3051,6 +3051,45 @@ This handles multi-line comments with a * prefix on each line."
...
+   ((and (boundp 'lsp-mode) lsp-mode)
+    (lsp-execute-code-action-by-kind kind))
+   (error "buffer is not managed by a known LSP client")))
+
+(defun go-freesymbols ()
+  "View free symbols referenced by the current selection. Requires gopls v0.16."
+  (interactive)
+  (go-mode--code-action "source.freesymbols"))
+
+(defun go-doc ()
```
dominikh, line 3072: Something like `go-browse-doc` might fit better, both to distinguish it from the existing `godoc` and to make it more apparent that it'll open the documentation in a browser. WDYT?

dominikh (approved):
Only a minor bikeshed, LGTM.

go-mode.el
adonovan: > Something like go-browse-doc might fit better, both to distinguish it from the existing godoc and to make it more apparent that it'll open the documentation in a browser. WDYT?

My preference was to omit it for the sake of brevity, but I confess I wasn't aware of the existing `godoc` function, so I now see the potential for confusion. Done.

adonovan:
(Don't merge this yet: I've come around to using the `go-browse-*` naming for all the verbs that open a browser, so I'll update the PR to fix `doc` and `assembly` too.)

adonovan:
> (Don't merge this yet: I've come around to using the `go-browse-*` naming for all the verbs that open a browser, so I'll update the PR to fix `doc` and `assembly` too.)

Done. Please take another look.

dominikh (approved):

adonovan:
We're planning to release gopls/v0.16 today, and its release notes will mention these new functions; it would be great if we could merge this PR soon. Do let me know if there's anything you need from me.

dominikh:
@adonovan Sorry for the delay. It's been merged to master now. Do you need a tagged release, or are people used to installing from master?

adonovan:
> @adonovan Sorry for the delay. It's been merged to master now. Do you need a tagged release, or are people used to installing from master?

No need to tag release; I have learned that Emacs users don't expect things to be easy. ;-)

Thanks.

PR #441, opened 2024-06-30
Update README.md to fix broken link to ELPA setup instructions

matta:


phikal:
Why not link to the Emacs manual: https://www.gnu.org/software/emacs/manual/html_node/emacs/Package-Installation.html?

